PR #6, opened 2016-03-10
Honor --bindir and --libdir options if present when building wave.

eldiener:
You should be able to tell the wave tool to build its libraries and/or executable outside of the Boost tree using the --libdir and/or --bindir command line options. The current build jamfile hardcodes the final lib and exe directory within the Boost tree. This fix allows the end-user to change that if he so desires.


hkaiser:
Thanks! LGTM!

PR #7, opened 2016-09-05
Preventing macro repetition.

Arash-codedev:
Macro repetition must be avoided to prevent clash with the other libraries.

A clash example is [here](https://github.com/Arash-codedev/Simple-Web-Server/blob/master/http_examples.cpp#L5)


hkaiser:
Thanks!

PR #8, opened 2016-10-05
Remove fs access, step 1

hia3:
Every called of include_paths::set_current_directory knows that path
argument is a name of a file (not a directory name).

Please don't apply this patch.


hkaiser:
What should I do with this PR if not merging it with the code base?

hia3:
This is just for illustration that every caller is passing `false` (meaning "not a directory") in all cases. You can disagree with some (all) of these cases. 

My plan was to replace some of these "false"s with a call to default_preprocessing_hooks::is_directory (to be able to redefine it later in client code). But none of them need this.

Close it :-)

PR #9, opened 2016-10-05
Remove fs access, step 2

hia3:


hkaiser:
I'd feel better if the conditional construct would be at least replaced by an `assert()` to make sure the change does not violate any preconditions for any of the users.

hkaiser:
Thanks!

PR #22, opened 2017-11-28
Repair C++11 raw string literal lexing

jefftrull:
Presently Wave expects C++11 raw string literals to be of the form `R"some text"` but this is not correct.  The format is actually `R"delim(some text)delim"`, where `delim` can be empty and `some text` may include anything other than `)delim`.

This PR adds code to the RE2C lexer to support the correct format, including checking delimiters for validity (permitted characters, length) and using them to quote arbitrary text which may include double quotes.  The raw string literal test is updated accordingly.

If merged, this PR will resolve [TRAC 12603](https://svn.boost.org/trac10/ticket/12603).


jefftrull:
I just realized it will also resolve [TRAC 12602](https://svn.boost.org/trac10/ticket/12602)

test/testwave/testfiles/t_7_001.cpp
```diff
@@ -11,29 +11,39 @@
...
+
+//R #line 45 "t_7_001.cpp"
+//R R"de"lim(some text)de"lim"
+R"de"lim(some text)de"lim"
+
+//R #line 49 "t_7_001.cpp"
 //R no_newline_at_end_of_file
-no_newline_at_end_of_file
\ No newline at end of file
+no_newline_at_end_of_file
```
hkaiser: This file is supposed to end without a trailing EOL, this accidentally adds it back in.

jefftrull:
Thanks for the heads up. My editor was apparently being "helpful".

PR #23, opened 2017-11-28
Add test/Jamfile

pdimov:
This adds a `Jamfile` in `test` as `wave` is the only library for which `b2 libs/libname/test` doesn't work. This would allow the elimination of the special case at https://github.com/boostorg/boost/blob/master/status/Jamfile.v2#L199.

### DIFF
--- a/test/Jamfile
+++ b/test/Jamfile
@@ -1,0 +1,3 @@
+import testing ;
+
+build-project build ;


pdimov:
Thanks. You can see the results of merging this in the Travis run on the superproject:

https://travis-ci.org/boostorg/boost/builds/308673108

Specifically, in the 8418.5 and 8418.6 jobs:

https://travis-ci.org/boostorg/boost/jobs/308673120
https://travis-ci.org/boostorg/boost/jobs/308673121

The Appveyor run of the same commit is at

https://ci.appveyor.com/project/boostorg/boost/build/1.0.4915-develop

It fails because, I suspect, the same happens there as in my local tests with MSVC, namely, `testwave.exe` throws up an assertion dialog box ("string iterators are incompatible") and the job times out. I haven't had time to investigate this.

hkaiser:
@jefftrull this looks like to be caused by your recent changes. Would you mind having a look?

jefftrull:
I'll get right on it.

pdimov:
I tried changing `s->act != Iterator()` to `s->act != s->last` and this got rid of the assertions, but tests started failing, so I assumed that it's not the right fix. On second look, however, the tests fail on`master` as well, so perhaps it's the right fix after all. The failures are of the form

```
C:\Projects\boost-git\boost\libs\wave\test\testwave\testfiles\t_6_029.cpp: failed
error result:
t_6_029.hpp(47): warning: unbalanced #if/#endif in include file: C:/Projects\\boost-git\\boost\\libs\\wave\\test\\testwave\\testfiles\\t_6_029.hpp

expected error:
t_6_029.hpp(47): warning: unbalanced #if/#endif in include file: C:\\Projects\\boost-git\\boost\\libs\\wave\\test\\testwave\\testfiles\\t_6_029.hpp

C:\Projects\boost-git\boost\libs\wave\test\testwave\testfiles\t_6_030.cpp: failed
error result:
t_6_030.hpp(49): warning: unbalanced #if/#endif in include file: C:/Projects\\boost-git\\boost\\libs\\wave\\test\\testwave\\testfiles\\t_6_030.hpp

expected error:
t_6_030.hpp(49): warning: unbalanced #if/#endif in include file: C:\\Projects\\boost-git\\boost\\libs\\wave\\test\\testwave\\testfiles\\t_6_030.hpp
```

That is, there's a mismatch between `C:/Projects` and `C:\\Projects` for some reason.

pdimov:
On msvc-8.0 - don't know if it's still supported - there are many more errors, of the form

```
C:\Projects\boost-git\boost\libs\wave\test\testwave\testfiles\t_9_023.cpp: failed
result:
#line zd "t_9_023.cpp"
_C_LIB_DECL

expected:
#line 18 "t_9_023.cpp"
_C_LIB_DECL

C:\Projects\boost-git\boost\libs\wave\test\testwave\testfiles\t_9_024.cpp: succeeded
testwave: 90 of 200 test(s) succeeded (110 test(s) failed).
```

That is, the `#line` directive shows `zd` instead of a line number.

pdimov:
Same `zd` errors on msvc-10.0, 11.0, 12.0. 14.0 and 14.1 no longer have them.

jefftrull:
Hi Peter, can you clarify whether the "zd" errors are observed when you apply the (s->act != s->last) hack, or without it (i.e., as the code currently stands)?  Thank you.

pdimov:
The `zd` errors are on the master branch, which doesn't need the change (which is not a hack :-) ).

jefftrull:
I withdraw the characterization :)

jefftrull:
I've reproduced the iterator complaints and will start debugging them soon.  The "C:\" vs "C://" issue appears to have been present as far back as 2012.  Haven't seen the "zd" thing yet but it may be masked by other problems.  Running MSVC 2017 now.

pdimov:
`zd` is only on 2013 and below. 2015/2017 don't have it.

jefftrull:
[This PR](https://github.com/boostorg/wave/pull/25) should resolve the iterator issue.  I am looking into some of the other problems (there seems to be a linker issue, also?)

jefftrull:
And [this](https://github.com/boostorg/wave/pull/26) should do the same for the linker errors

jefftrull:
The `unbalanced #if/#endif` miscompares are happening because Boost.Filesystem's path normalization function changes the root path separator.  See [here](https://stackoverflow.com/questions/27110973/boost-filesystem-canonical-path-is-not-valid-after-conversion-to-const-char) for a detailed discussion. I applied the suggested workaround in [this PR](https://github.com/boostorg/wave/pull/27).

jefftrull:
It looks like the "zd" is coming from [here](https://github.com/boostorg/wave/blob/develop/include/boost/wave/util/cpp_iterator.hpp#L812) - perhaps sprintf is broken in msvc-12.0 for the "%zd" format?

pdimov:
`%zd` is probably not supported on msvc-12.0 or below.

pdimov:
https://msdn.microsoft.com/en-us/library/tcxf1dw6.aspx

"The hh, j, z, and t length prefixes are not supported."

pdimov:
It appears to work on msvc-14.0 though, despite what the doc says.

pdimov:
The 2017 docs list it as supported: https://docs.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions#size

pdimov:
Appveyor is green now: https://ci.appveyor.com/project/boostorg/boost/build/1.0.4932-develop
The matrix http://www.boost.org/development/tests/develop/developer/wave.html will also turn green in a few days. Even the `__boost_check_library__` row has been fixed, by this PR in combination with https://github.com/boostorg/boost/commit/3af09d83695bc1dda0567d0cc7570f3639bda59e.

Good work, us. :-)

pdimov:
Would it be possible to have this commit merged to master? I want to merge the superproject change.

hkaiser:
Uhh, I thought I had merged it... Sorry. Will do.

hkaiser:
Done. Thanks!

PR #25, opened 2017-11-30
Remove comparisons vs. default-initialized (and thus singular) iterators

jefftrull:
MSVC's standard library has checked iterators which flagged that we compare default-initialized iterators against regular iterators, which is not allowed by the standard.  The purpose of those comparisons is to ensure the iterators (formerly strictly pointers) are initialized before dereferencing them.
As it turns out, a constructor was added to the Scanner class as part of the changes that introduced this problem, and so the iterators will always be initialized, and we can remove these checks.


pdimov:
Hartmut, would you like me to add Travis support to this repo, so that the pull requests are tested?

PR #26, opened 2017-11-30
Resolve link errors in MSVC builds by exporting symbols

jefftrull:
In the forward iterator PR a number of functions became function templates and so moved into .hpp files. As a result some symbols were not locally visible to their users anymore, and now need to be exported for MSVC builds.

### DIFF
--- a/include/boost/wave/cpplexer/re2clex/cpp_re.hpp
+++ b/include/boost/wave/cpplexer/re2clex/cpp_re.hpp
@@ -173,7 +173,7 @@ int count_backslash_newlines(Scanner<Iterator> *s, uchar *cursor)
     return skipped;
 }
 
-bool is_backslash(uchar *p, uchar *end, int &len);
+BOOST_WAVE_DECL bool is_backslash(uchar *p, uchar *end, int &len);
 
 #define BOOST_WAVE_BSIZE     196608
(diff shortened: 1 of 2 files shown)


include/boost/wave/cpplexer/re2clex/cpp_re.hpp
```diff
@@ -350,18 +350,18 @@ uchar *fill(Scanner<Iterator> *s, uchar *cursor)
...
 
-    operator uchar *() const;
+    BOOST_WAVE_DECL operator uchar *() const;
 
     friend std::ptrdiff_t
-    operator- (uchar_wrapper const& lhs, uchar_wrapper const& rhs);
+    BOOST_WAVE_DECL operator- (uchar_wrapper const& lhs, uchar_wrapper const& rhs);
 
     uchar *base_cursor;
     std::size_t column;
```
hkaiser: It might be better to add the `BOOST_WAVE_DECL` to the whole class (`struct BOOST_WAVE_DECL uchar_wrapper {...};`) then to every single member functions.

jefftrull: Ah.... you can do that? Fixing :)

PR #29, opened 2017-11-30
Fixing cpp11 format specs

jefftrull:
Wave uses the "%zd" format spec in a couple of places.  Unfortunately, the "z" modifier was introduced in C+11 and so it does not work with old versions of MSVC.  This commit replaces `sprintf` with `boost::lexical_cast`.

### DIFF
--- a/include/boost/wave/util/cpp_iterator.hpp
+++ b/include/boost/wave/util/cpp_iterator.hpp
@@ -796,7 +796,7 @@ typename ContextT::position_type pos = act_token.get_position();
 
             if (!ctx.get_hooks().emit_line_directive(ctx, pending, act_token))
             {
-            unsigned int column = 6;
+                unsigned int column = 6;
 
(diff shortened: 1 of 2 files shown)


include/boost/wave/util/cpp_iterator.hpp
```diff
@@ -806,14 +807,12 @@ typename ContextT::position_type pos = act_token.get_position();
 
             // 21 is the max required size for a 64 bit integer represented as a
             // string
-            char buffer[22];
 
-                using namespace std;    // for some systems sprintf is in namespace std
-                sprintf (buffer, "%zd", pos.get_line());
+		std::string buffer = lexical_cast<std::string>(pos.get_line());
```
hkaiser: The indentation got messed up somehow.

jefftrull: That whole region is strange.... I'll clean it up.

Issue #32, opened 2018-09-28
wave can't process includes with unicode characters in paths on windows


rlenhardt:
Suppose i have a directory ßµ containing header.h with content

    const char* file() { return __FILE__; }

and source.cpp with content

    #include <header.h>
    #include <iostream>
    int main() { std::cout << file() << std::endl; }

Now on windows calling the wave driver tool in the directory ßµ like this
`wave -I. source.cpp`
fails to find the header, same when calling wave from the parent directory of ßµ:
`wave -Ißµ ßµ\source.cpp`
Moving source.cpp to the parent directory will fail when trying to open the included header.
Changing the include in source.cpp to `#include <ßµ/header.h>` and calling wave with
`wave -I. source.cpp`
also fails to include the header.

All this stems from wave using 8 bit strings for filenames and include paths internally and using standard streams to open and read the actual files. The conversion to boost::filesystem::path and back to string seems to mess up the path encoding at some place. This seems to be a problem for many years, see https://svn.boost.org/trac10/ticket/6970.

A quick fix would be to use UTF-8 as path encoding in wave by setting up boost::filesystem::path to recognize UTF-8, this can be done at that start of wave or outside of wave like this (taken from [here](http://boost.2283326.n4.nabble.com/boost-filesystem-path-as-utf-8-tp4320098p4322460.html)):

    #include <boost/filesystem/detail/utf8_codecvt_facet.hpp>
    ...
    std::locale global_loc = std::locale();
    std::locale loc(global_loc, new
    boost::filesystem::detail::utf8_codecvt_facet);
    boost::filesystem::path::imbue(loc); 

To make this work, all standard streams used to open files have to be switched to the boost filesystem stream wrappers. Then the 8 bit string filename input of the stream open method will be taken as UTF-8 encoded and opened correctly.

I have tried that this approach works. While you could easily adapt the Input Policy of wave from outside of wave it would be nice if wave and the wave driver would use the boost filesystem wrappers internally, so changing the boost filesystem locale like mentioned above just works.

If this sounds like a good idea to include in the wave lib i can set up a pull request.

Thanks

hkaiser:
@rlenhardt I'd appreciate a corresponding PR! Thanks!

PR #33, opened 2018-09-28
use boost filesystem streams for opening files in wave

rlenhardt:
This is the PR for issue #32.
The first commit changes standard streams to boost streams in the wave lib and the wave driver. I left out the samples.
The second submit is a demo on how to switch the wave driver to UTF-8 paths. I had no luck passing UTF-8 arguments to wave using cmd.exe, you have to switch to mintty or something. But all invocations of wave without passing in UTF-8 paths work. I'd rather not pull the second submit as-is, maybe the locale switch should be optional or left away altogether. I would be happy if the first commit made it into boost wave.


rlenhardt:
Will close this PR and open another one, without the DEMO code but with tests.

PR #37, opened 2018-10-04
use boost filesystem streams in wave to enable utf-8 paths in includes

rlenhardt:
This is the PR for issue #32.
The first commit changes standard streams to boost streams in the wave lib and the wave driver. I left out the samples.
The second submit adds 4 tests, two for checking that include paths containing UTF-8 characters work in include directives and passed via commandline include paths, the two other check that those characters fail when not switching to the boost filesystem utf8 codecvt -- at least on windows, they might not fail on OSs that always use UTF8 for path encoding.

I added a new flag for the special directives to switch to the UTF8 codecvt on a testcase basis. I hope the special include dir does not cause problems when checking out via git.


test/testwave/testwave_app.cpp
```diff
@@ -1009,11 +1041,14 @@ testwave_app::initialise_options(Context& ctx, po::variables_map const& vm,
         for (std::vector<std::string>::const_iterator cit = syspaths.begin();
               cit != end; ++cit)
         {
+            std::string full(*cit);
+            got_expected_result(ctx.get_current_filename(),"",full);
```
rlenhardt: I (ab)used the got_expected_results function to make the include paths into full paths via the $P placeholder.

hkaiser: Fair enough.

hkaiser:
Things are failing: https://travis-ci.org/boostorg/wave/jobs/437812242#L1039. Could you have a look, please? We could enable the UTF8 changes conditionally...

pdimov:
As we suspected, UTF-8 paths in the repo are proving problematic, see https://github.com/boostorg/boostdep/issues/8. I'm fixing that by putting the path into a .zip file and extracting it in .travis.yml and appveyor.yml. This solution won't work here though as it's part of the test suite.

PR #45, opened 2019-05-03
Link against Boost.Filesystem to resolve missing symbols in samples

jefftrull:
I noticed several samples would fail at the link stage, all with missing Boost.Filesystem symbols. I couldn't find a way to have them inlined so I assume this changed at some point. Adding FS explicitly fixes it.


jefftrull:
Possibly related to [TRAC 6589](https://svn.boost.org/trac10/ticket/6589)

PR #46, opened 2019-05-03
Instantiate lexers with std::string, not flex_string

jefftrull:
Two samples fail at the link stage looking for a different `new_lexer_gen<>::new_lexer()` than we provide.

AFAICT the type of the lexer is dependent on what string was used to construct the wave context object and that propagates down. Updating these files fixes the error.


jefftrull:
I think this resolves [TRAC 6588](https://svn.boost.org/trac10/ticket/6588)

PR #57, opened 2020-01-13
Keep develop up-to-date with master

jefftrull:
Looks like we've missed a single one-line bugfix from 2015 that went straight into master, from @joshguffin "Fix off-by-one error". Nevertheless it lists 169 commits, which is strange...


jefftrull:
We could perhaps cherry-pick that one commit into develop and let the history stay as it is.

hkaiser:
Yah, cherry picking looks like the sensible option...

jefftrull:
So it shall be.

Issue #64, opened 2020-02-10
Invalid C++03 syntax in sample


jefftrull:
A recent change introduced the following line into `real_position_token.hpp`:

`using boost::wave::token_id::T_UNKNOWN;`

This is C++11 syntax and will break CI if samples are built there.

PR #65, opened 2020-02-10
Remove alias for T_UNKNOWN and just use boost::wave::T_UNKNOWN

jefftrull:


jefftrull:
If merged, this will fix #64

PR #66, opened 2020-02-11
Ensure wave driver and sample programs build in CI, fixing issues turned up during the process.

### LINKED ISSUE #48: Samples should be built by CI
This will keep the sample programs healthy over time. They don't seem to be built by Travis. If the recent fixes work they can maybe be kept buildable this way.

jefftrull:
Display a Travis build status badge and some minimal description in a README

If merged this will resolve #48.

### DIFF
--- a/samples/token_statistics/build/Jamfile.v2
+++ b/samples/token_statistics/build/Jamfile.v2
@@ -19,7 +19,7 @@ exe token_statistics
     :
         $(SOURCES)
         /boost/wave//boost_wave
-        /boost/program_options//boost_program_options
+        /boost/program_options//boost_program_options/<link>static
         /boost/filesystem//boost_filesystem     
         /boost/system//boost_system     
         /boost/thread//boost_thread
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,12 +75,14 @@ install:
   - git submodule update --init libs/config
   - git submodule update --init tools/boostdep
   - cp -r $TRAVIS_BUILD_DIR/* libs/wave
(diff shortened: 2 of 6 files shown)


samples/token_statistics/build/Jamfile.v2
```diff
@@ -19,7 +19,7 @@ exe token_statistics
     :
         $(SOURCES)
         /boost/wave//boost_wave
-        /boost/program_options//boost_program_options
+        /boost/program_options//boost_program_options/<link>static
```
hkaiser: What's the rationale of this change? Could you elaborate, please?

jefftrull: Sure! This stems from a problem linking on MSVC-12. A nearly identical issue was reported at https://github.com/boostorg/program_options/issues/58; after investigating it for a while I realized that the tactic used [here](https://github.com/boostorg/wave/blob/18f834ac355d65d18619af673122c23623e7c912/test/build/Jamfile.v2#L93) would fix this as well.

I did also try adding `BOOST_ALL_DYN_LINK` but then it seemed like some inappropriate things got marked `__dllimport`, in particular `expression_grammar_gen<TokenT>::evaluate` and MSVC didn't like that either.

I haven't tried using `BOOST_PROGRAM_OPTIONS_DYN_LINK` but could attempt it if you prefer that approach.

hkaiser: Don't worry about this too much, I was just curious.

Issue #54, opened 2020-01-11
turning off 'include_next' support breaks includes


jefftrull:
This is a port of [TRAC 9874](https://svn.boost.org/trac10/ticket/9874). The original text reads:

> The directive '#if BOOST_WAVE_SUPPORT_INCLUDE_NEXT !+ 0" in file cpp_re2_lexer.hpp is in the wrong place, and causes all #include processing to break if BOOST_WAVE_SUPPORT_INCLUDE_NEXT is set to 0.
> 
> I've attached a simple patch.
[wave.2.patch](https://github.com/boostorg/wave/files/4048939/wave.2.patch.txt)

jefftrull:
I (@jefftrull ) reproduced this problem and also verified the patch fixes it. The attached test code produces the message `caught exception: error: ill formed #include directive: #include` when `BOOST_WAVE_SUPPORT_INCLUDE_NEXT` is defined as 0.
[trac9874.cpp](https://github.com/boostorg/wave/files/4048941/trac9874.cpp.txt)

jefftrull:
Closing thanks to merged fix

PR #67, opened 2020-02-11
Fix the problem where disabling `include_next` support breaks regular `include`.

jefftrull:
It applies the same fix suggested by the original poster of #54 and if merged will resolve it.


jefftrull:
Hm, I see I didn't squash this one. I can if you like.

PR #68, opened 2020-02-12
Restore waveidl sample to working condition

### LINKED ISSUE #56: waveidl sample does not use the IDL lexer, but the default lexer
This is a port of [TRAC 7822](https://svn.boost.org/trac10/ticket/7822). The original text begins:
> Dear Boost developpers,
>
> The waveidl sample from the wave lib does not use the included IDL lexer. Instead it uses the default cpp lexer. This can be easily tested by setting a breakpoint in the scan function in idl_re.cpp. The reason is that it is explicitly set up like this in the lex_iterator constructor.

jefftrull:
- use the idllexer generator instead of the cpp one
- eliminate a double free encountered during fix

If merged, this will resolve #56

### DIFF
--- a/samples/waveidl/idllexer/idl_re2c_lexer.hpp
+++ b/samples/waveidl/idllexer/idl_re2c_lexer.hpp
@@ -116,7 +116,6 @@ template <typename IteratorT, typename PositionT>
 inline
 lexer<IteratorT, PositionT>::~lexer() 
 {
-    boost::wave::cpplexer::re2clex::aq_terminate(scanner.eol_offsets);
     free(scanner.bot);
 }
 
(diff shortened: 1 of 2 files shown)


samples/waveidl/idllexer/idl_re2c_lexer.hpp
```diff
@@ -116,7 +116,6 @@ template <typename IteratorT, typename PositionT>
 inline
 lexer<IteratorT, PositionT>::~lexer() 
 {
-    boost::wave::cpplexer::re2clex::aq_terminate(scanner.eol_offsets);
```
hkaiser: I don't really remember why `aq_terminate` was in here. Is it not needed anymore?

jefftrull: It is some kind of cleanup code that I don't understand, but what I do know is it already gets called from scanner's destructor :)

jefftrull: Also the equivalent call is not present in cpplexer.

hkaiser: Ok, thanks for fixing this!

Issue #61, opened 2020-01-22
wave driver option --traceto


quinzio:
Hello to everyone, 
I am very new to boost library, so please excuse me if my supposed issue is due to my lack of experience.

I am using the boost wave driver example, that you find in the tools folder of the library. 
The --traceto option does not produce any output, even if I specify a file, or to the std err ( - parameter). 
My example is very basic and obvious. 
Thanks.

wave --traceto - myfile.cpp

myfile.cpp:

#define A 1
#define B 3
#define C (A + B)
#define D (2 * C)

int main() {
    int k = D;
    return C;
}

jefftrull:
Thanks for the feedback! It looks like a bug to me.

@hkaiser I think maybe this center operator should be an `||` instead of `&&` [here](https://github.com/boostorg/wave/blob/69239bba69f4e78ebb09c1c8abe6118f5735f2ed/tool/trace_macro_expansion.hpp#L1405), what do you think? When I make that change both `-c` and `-t` options work again.

@quinzio give that fix a try and see if it helps you.

quinzio:
> Thanks for the feedback! It looks like a bug to me.
> 
> @quinzio give that fix a try and see if it helps you.

Yep, now it's working and macros expansions are pouring down.
I changed the line you mentioned from
        return (flags & trace_macros) && (logging_flags & trace_macros);
to
        return (flags & trace_macros) || (logging_flags & trace_macros);

Thanks for the quick reply. Wave is a useful tool to navigate into heaviliy preprocessed software.

hkaiser:
I think the initial idea was that you would have to enable the tracing from inside your code using a wave specific `#pragma` (or `_Pragma()`) in addition to specifying the command line flag. This was done in order to be able to limit the generated trace to the macro(s) of interest. Otherwise the information produced tends to be overwhelming. Please see [here](https://www.boost.org/doc/libs/1_72_0/libs/wave/doc/supported_pragmas.html) for details.

jefftrull:
Thanks @hkaiser I do see that now. I was also confused though :) Shall we call it a documentation bug? I think adding some information to [the invocation page](https://www.boost.org/doc/libs/1_72_0/libs/wave/doc/wave_driver.html) could help.

hkaiser:
@jefftrull yes, I agree. Let's rather fix the documentation. Thanks!

jefftrull:
[...]

PR #69, opened 2020-02-16
Add a note to invocation page explaining about trace(enable)

jefftrull:
IMO this will resolve the --traceto usage confusion (at least, it might have resolved mine) and thus #61

### DIFF
--- a/doc/wave_driver.html
+++ b/doc/wave_driver.html
@@ -170,7 +170,7 @@
     log. If the filename given equals to <tt>'-'</tt> (without the quotes), the 
     trace log is put into the standard error stream (stderr).</p>
   <p dir="ltr">Tracing is only performed for portions of the input where it has been
-    explicitly enabled via the <tt>#pragma wave trace(enable)</tt> directive.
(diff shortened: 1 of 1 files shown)


doc/wave_driver.html
```diff
@@ -165,10 +165,13 @@
 </blockquote>
 <p dir="ltr">-t [--traceto] path</p>
 <blockquote> 
-  <p dir="ltr">Enable the tracing facility build into the <tt>Wave</tt> library. 
+  <p dir="ltr">Enable the tracing facility built into the <tt>Wave</tt> library.
     The path specifies the filename to use for the output of the generated trace 
     log. If the filename given equals to <tt>'-'</tt> (without the quotes), the 
     trace log is put into the standard error stream (stderr).</p>
+  <p dir="ltr">Tracing is only performed for portions of the input where it has been
+    explicitly enabled via the <tt>#pragma wave trace(enable)</tt> directive.
```
hkaiser: Perhaps we should explicitly mention the inline `_Pragma` version as it's more useful for this kind of thing:
```suggestion
    explicitly enabled via the <tt>#pragma wave trace(enable)</tt> or <tt>_Pragma("wave trace(enable)")</tt> directive.
```

PR #75, opened 2020-02-28
Implement C++20 features

jefftrull:
I believe this implements the two key features:

1. the ability to supply 0 variadic arguments in more cases
2. using `__VA_OPT__` in macro definitions

however it's a big change and I'm not sure I did it right. Hoping for some feedback and very happy to change things around.


jefftrull:
This is targeting 1.74, if that wasn't clear

include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -1233,9 +1302,21 @@ ContainerT replacement_list;
...
+#if BOOST_WAVE_SUPPORT_CPP2A
+        if (boost::wave::need_cpp2a(ctx.get_language())) {
+            // if the last parameter is ellipsis (varargs),
+            // reduce the required count by one
+            if ((parm_count_required > 0) &&
+                (T_ELLIPSIS == token_id(macro_def.macroparameters.back()))) {
+                --parm_count_required;
+            }
+        }
+#endif
```
hkaiser: Would you mind adding a comment explaining that in C++20 the rules have changed for how many arguments are required if ellipses are involved? I.e. something like:
> The number of arguments must be more than (until C++20) / at least as many as (since C++20) the number of parameters (not counting ...).

include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -1004,13 +1023,56 @@ bool adjacent_stringize = false;
...
+                            improperly_terminated_macro, "missing ')' in __VA_OPT__",
+                            pos);
+                    }
+                    // cstart still points to __VA_OPT__; cit now points to the last rparen
+
+                    if ((i == arguments.size()) ||             // no variadic argument
+                        impl::is_blank_only(arguments[i])) {   // no visible tokens
+                        // no args; insert placemarker
+                        *std::back_inserter(expanded) =
+                            typename ContainerT::value_type(T_PLACEMARKER, "\xA7", pos);
```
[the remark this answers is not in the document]
hkaiser: Also, this inserts a place marker even if the token sequence already contains one. Is that intended?

include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -1046,12 +1108,19 @@ bool adjacent_stringize = false;
...
+                        impl::trim_sequence_left(arguments[i]);
+                        impl::trim_sequence_right(arguments.back());
+                        BOOST_ASSERT(boost::wave::need_variadics(ctx.get_language()));
+                        impl::replace_ellipsis(arguments, i, expanded, pos);
+                    } else {
+                        BOOST_ASSERT(i == arguments.size());
+                        // no argument supplied; insert placemarker
+                        *std::back_inserter(expanded) =
+                            typename ContainerT::value_type(T_PLACEMARKER, "\xA7", pos);
+                    }
```
hkaiser: Should this `else` be C++20 only? Or does it fix a bug?

include/boost/wave/util/macro_helpers.hpp
```diff
@@ -252,6 +252,48 @@ namespace impl {
...
+            if (T_RIGHTPAREN == token_id(*mdit)) {
+                scope--;
+            } else if (T_LEFTPAREN == token_id(*mdit)) {
+                scope++;
+            }
+        }
+        if ((mdit == mdend) && ((scope != 1) || (T_RIGHTPAREN != token_id(*mdit)))) {
+            // arrived at end without matching rparen
+            mdit = mdstart_it;
+            return false;
```
hkaiser: Should this rather report an error?

test/testwave/testfiles/t_8_001.cpp
```diff
@@ -0,0 +1,44 @@
...
+SDEF(foo);
+SDEF(bar, 1, 2, 3);
+
+//R #line 43 "t_8_001.cpp"
+//R printf("at line=%d" , 43); printf("\n");
+// P00306 example suggests the adjacent string literals are concatenated in
+// this test case but that's the *compiler's* job:
+//R printf("at line=%d" ": ", 44); printf("All well in zone %d", n); printf("\n");
+LOG2();
+LOG2("All well in zone %d", n);
```
hkaiser: Do you think we should add the hook verification as well (i.e. `//H...` comments)?

tool/cpp.cpp
```diff
@@ -1329,6 +1349,9 @@ main (int argc, char *argv[])
 #endif
 #if BOOST_WAVE_SUPPORT_CPP0X != 0
             ("c++11", "enable C++11 mode (implies --variadics and --long_long)")
+#endif
+#if BOOST_WAVE_SUPPORT_CPP2A != 0
+        ("c++20", "enable C++20 mode (implies --variadics and --long_long, adds __VAR__OPT__)")
```
hkaiser: Indentation?

test/testwave/testfiles/t_8_001.cpp
hkaiser: FWIW, adding a `//H` line at the end of t_8_001.cpp makes testwave produce the following recorded hooks:
```c++
//H 10: t_8_001.cpp(17) : #define
//H 08: t_8_001.cpp(17) : LOG(msg, ...) = printf(msg __VA_OPT__(, ) __VA_ARGS__)
//H 10: t_8_001.cpp(18) : #define
//H 08: t_8_001.cpp(18) : SDEF(sname, ...) = S sname __VA_OPT__(= { __VA_ARGS__ })
//H 10: t_8_001.cpp(19) : #define
//H 08: t_8_001.cpp(19) : LOG2(...) = printf("at line=%d" __VA_OPT__(": "), __LINE__);      __VA_OPT__(printf(__VA_ARGS__);)                      printf("\n")
//H 00: t_8_001.cpp(28) : LOG("hello world\n"), [t_8_001.cpp(17):LOG(msg, ...) = printf(msg __VA_OPT__(, ) __VA_ARGS__)]
//H 02: printf("hello world\n" º)
//H 03: printf("hello world\n")
//H 00: t_8_001.cpp(29) : LOG("hello world\n", º), [t_8_001.cpp(17):LOG(msg, ...) = printf(msg __VA_OPT__(, ) __VA_ARGS__)]
//H 02: printf("hello world\n" º)
//H 03: printf("hello world\n")
//H 00: t_8_001.cpp(30) : LOG("hello %d\n", n), [t_8_001.cpp(17):LOG(msg, ...) = printf(msg __VA_OPT__(, ) __VA_ARGS__)]
//H 02: printf("hello %d\n", n)
//H 03: printf("hello %d\n", n)
//H 00: t_8_001.cpp(35) : SDEF(foo), [t_8_001.cpp(18):SDEF(sname, ...) = S sname __VA_OPT__(= { __VA_ARGS__ })]
//H 02: S foo º
//H 03: S foo
//H 00: t_8_001.cpp(36) : SDEF(bar, 1, 2, 3), [t_8_001.cpp(18):SDEF(sname, ...) = S sname __VA_OPT__(= { __VA_ARGS__ })]
//H 02: S bar = { 1, 2, 3 }
//H 03: S bar = { 1, 2, 3 }
//H 00: t_8_001.cpp(43) : LOG2(º), [t_8_001.cpp(19):LOG2(...) = printf("at line=%d" __VA_OPT__(": "), __LINE__);      __VA_OPT__(printf(__VA_ARGS__);)                      printf("\n")]
//H 02: printf("at line=%d" º, __LINE__);      º                      printf("\n")
//H 03: printf("at line=%d", 43);                            printf("\n")
//H 00: t_8_001.cpp(44) : LOG2("All well in zone %d", n), [t_8_001.cpp(19):LOG2(...) = printf("at line=%d" __VA_OPT__(": "), __LINE__);      __VA_OPT__(printf(__VA_ARGS__);)                      printf("\n")]
//H 02: printf("at line=%d" ": ", __LINE__);      printf("All well in zone %d", n);                      printf("\n")
//H 03: printf("at line=%d" ": ", 44);      printf("All well in zone %d", n);                      printf("\n")
```
This looks bogus as it apparently contains placemarker tokens (the `º` characters). I think that shouldn't happen.

include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -1004,13 +1023,56 @@ bool adjacent_stringize = false;
...
+                    // cstart still points to __VA_OPT__; cit now points to the last rparen
+
+                    if ((i == arguments.size()) ||             // no variadic argument
+                        impl::is_blank_only(arguments[i])) {   // no visible tokens
+                        // no args; insert placemarker
+                        *std::back_inserter(expanded) =
+                            typename ContainerT::value_type(T_PLACEMARKER, "\xA7", pos);
+                    } else {
+                        ++cstart; ++cstart;
```
hkaiser: I think we should make sure that the replacement list does not contain a `__VA_OPT__` itself (P0306r4 states:
> The identifier __VA_OPT__ shall always occur as part of the token sequence __VA_OPT__(content), where content is an arbitrary sequence of preprocessing-tokens other than __VA_OPT__.

include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -1004,13 +1023,56 @@ bool adjacent_stringize = false;
                     impl::replace_ellipsis(expanded_args, i, expanded, pos);
                 }
                 else
+
+#if BOOST_WAVE_SUPPORT_VA_OPT != 0
+                if (is_va_opt) {
```
hkaiser: I using `__VA_OPT__` without ellipses in the macro definition parameter list valid?

include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -1004,13 +1023,56 @@ bool adjacent_stringize = false;
...
+                        // no args; insert placemarker
+                        *std::back_inserter(expanded) =
+                            typename ContainerT::value_type(T_PLACEMARKER, "\xA7", pos);
+                    } else {
+                        ++cstart; ++cstart;
+                        // [cstart, cit) is now the args to va_opt
+                        // recursively process them
+                        expand_replacement_list(cstart, cit, arguments,
+                                                expand_operator_defined, expanded);
+                    }
```
hkaiser: I'm not sure if recursively expanding the replacement list is the correct solution here. I think calling `impl::replace_ellipsis` (like for `__VA_ARGS__` above) in connection with the rescanning that is done afterwards  should be sufficient. Doing recursive expanding of the replacement list might actually expand too much.

include/boost/wave/util/cpp_macromap_utils.hpp
```diff
@@ -343,6 +343,27 @@ is_whitespace_only (ContainerT const &argument)
...
 
+///////////////////////////////////////////////////////////////////////////////
+//
+//  Tests whether the given token sequence consists only of whitespace
+//  and placemarkers
+//
+///////////////////////////////////////////////////////////////////////////////
+template <typename ContainerT>
+inline bool
+is_blank_only (ContainerT const &argument)
```
jefftrull: @hkaiser would love your thoughts here. The purpose of this function is to help detect "no variadic argument supplied" cases. I began with
1. ellipsis index == argument count (so no argument present)

but found I had to add

2. argument consists only of whitespace (present but empty?)

and then modify it to

2. argument consists only of whitespace or placemarkers (thus "is blank")

to get it to work, and that just seems hacky.

hkaiser: @jefftrull I understand. IIUC, this function is currently used in a context where you want to decide whether a placemarker has to be inserted or not. However, `is_blank_only` will return `true` even if there already is a placemarker, causing for another one to be inserted into the replacement list. 

I think using `is_whitespace_only` in this context would be sufficient as you want to insert a placemarker whenever there isn't anything but whitespace in the processed sequence.

test/testwave/testfiles/t_8_001.cpp
hkaiser: Actually, the placemarker symbols may only become removed after the `H 02` (`expanded_macro` hook) is called (during rescanning of the replacement list). Thus, seeing those in the output could be correct after all. I just have never seen any placemarkers being printed in this context before. 

We might want to remove placemarkers from the generated string representation, similarly to what we do while generating string literals (see: https://github.com/boostorg/wave/blob/develop/include/boost/wave/util/macro_helpers.hpp#L128-L135).

doc/wave_driver.html
```diff
@@ -56,7 +56,8 @@
     --long_long:                 enable long long support if C++ mode
     --variadics:                 enable variadics and placemarkers in C++ mode
     --c99:                       enable C99 mode (implies variadics and placemarkers)
-    --c++11                      enable C++11 mode (implies --variadics and --long_long)
+    --c++11:                     enable C++11 mode (implies --variadics and --long_long)
+    --c++20:                     enable C++20 mode (adds __VA_OPT__ to variadics)
```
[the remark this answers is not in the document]
jefftrull: I guess it does :)

include/boost/wave/language_support.hpp
```diff
@@ -35,12 +35,21 @@ enum language_support {
...
     support_cpp11 = support_cpp0x,
 #endif
+#if BOOST_WAVE_SUPPORT_CPP2A != 0
+//  support flags for C++20
+    support_option_va_opt = 0x10000,
+
+    support_cpp2a = support_option_variadics | support_option_long_long |
+        support_option_no_newline_at_end_of_file | support_option_va_opt | 0x20000,
+    support_cpp20 = support_cpp2a,
+#endif
```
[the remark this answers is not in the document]
jefftrull: Yes, given that it depends on `support_option_no_newline_at_end_of_file`! I will make the change.

include/boost/wave/token_ids.hpp
```diff
@@ -311,7 +312,8 @@ enum token_id {
     T_PLACEHOLDER = TOKEN_FROM_ID(T_LAST_TOKEN+2, WhiteSpaceTokenType),
     T_PLACEMARKER = TOKEN_FROM_ID(T_LAST_TOKEN+3, InternalTokenType),
     T_PARAMETERBASE = TOKEN_FROM_ID(T_LAST_TOKEN+4, ParameterTokenType),
-    T_EXTPARAMETERBASE = TOKEN_FROM_ID(T_LAST_TOKEN+4, ExtParameterTokenType)
+    T_EXTPARAMETERBASE = TOKEN_FROM_ID(T_LAST_TOKEN+4, ExtParameterTokenType),
+    T_OPTPARAMETERBASE = TOKEN_FROM_ID(T_LAST_TOKEN+4, OptParameterTokenType)
```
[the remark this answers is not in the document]
jefftrull: I am following the pattern of `T_PARAMETERBASE` AND `T_EXTPARAMETERBASE` here - both have the same id but a different type. So if `T_EXTPARAMETERBASE` is correct this should be as well, I think (hope?)

include/boost/wave/util/cpp_iterator.hpp
```diff
@@ -1806,6 +1818,11 @@ position_type pos(act_token.get_position());
                     {
                         seen_va_args = true;
                     }
+                    if (T_IDENTIFIER == token_id(*dit) &&
+                        "__VA_OPT__" == (*dit).get_value())
+                    {
+                        seen_va_opt = true;
+                    }
```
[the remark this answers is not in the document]
jefftrull: I will do both of those as well as the declaration of `seen_va_opt`

include/boost/wave/util/cpp_iterator.hpp
```diff
@@ -1814,6 +1831,12 @@ position_type pos(act_token.get_position());
                         macroname.get_value().c_str(), act_token.get_position());
                     return;
                 }
+                if (seen_va_opt) {
+                    BOOST_WAVE_THROW_CTX(ctx, preprocess_exception,
+                        bad_define_statement_va_opt,
+                        macroname.get_value().c_str(), act_token.get_position());
+                    return;
+                }
```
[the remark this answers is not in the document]
jefftrull: I will do both of those as well as the declaration of `seen_va_opt`

include/boost/wave/token_ids.hpp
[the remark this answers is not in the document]
hkaiser: Fair enough!

tool/cpp.cpp
hkaiser: Also, should probably be `__VAR_OPT__`.

include/boost/wave/util/cpp_macromap.hpp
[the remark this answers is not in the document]
jefftrull: 1. Yes :) I will make the change.
2. I'm not sure how to avoid inserting two placemarkers. Is there an appropriate Wave idiom, or a different technique I should use?

include/boost/wave/util/cpp_macromap_predef.hpp
```diff
@@ -240,6 +240,23 @@ namespace util {
...
+#if BOOST_WAVE_SUPPORT_CPP2A != 0
+    // C++20 mode
+        static_macros const& static_data_cpp2a(std::size_t i) const
+        {
+        static static_macros data[] = {
+                { "__STDC__", T_INTLIT, "1" },
+                { "__cplusplus", T_INTLIT, "202002L" },
+                { "__STDC_VERSION__", T_INTLIT, "199901L" },
+                { "__STDC_HOSTED__", T_INTLIT, "0" },
+                { "__WAVE_HAS_VARIADICS__", T_INTLIT, "1" },
```
hkaiser: Should we define a `{ "__WAVE_HAS_VA_OPT__", T_INTLIT, "1" },` to enable for user code to detect support for it? I'm on the fence here whether this is necessary...

include/boost/wave/util/cpp_macromap.hpp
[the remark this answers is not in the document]
hkaiser: > 2. I'm not sure how to avoid inserting two placemarkers. Is there an appropriate Wave idiom, or a different technique I should use?

I think using `is_whitespace_only` instead of `is_blank_only` would be sufficient to resolve this.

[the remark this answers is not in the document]
jefftrull: Well, unfortunately making that change causes two of my test cases to fail: [this one](https://github.com/jefftrull/wave/blob/9121d11e20adb392160f2d4ec8f19a64a901677b/test/testwave/testfiles/t_8_001.cpp#L29) and [this one](https://github.com/jefftrull/wave/blob/9121d11e20adb392160f2d4ec8f19a64a901677b/test/testwave/testfiles/t_8_001.cpp#L43). And that's why the code was change to test for placemarkers... Why that was necessary I'm not sure and I would not be surprised if there was a better solution (or a bug I introduced elsewhere).

include/boost/wave/util/cpp_iterator.hpp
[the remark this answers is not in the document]
hkaiser: P0306r4 introduces `__VA_OPT__` as a _functional_ macro. That means that it has to be followed by a `T_LEFT_PAREN` (possibly with whitespace in between) in order to be recognized as such. Using `__VA_OPT__` without a parenthesis should not invoke the corresponding logic but should leave the token alone, possibly for later expansion.

include/boost/wave/util/cpp_macromap.hpp
jefftrull: Sounds reasonable. I will make the change.

include/boost/wave/util/cpp_macromap.hpp
hkaiser: P0306r4 states that `__VA_OPT__` may only be used in the replacement text of a variadic macro.

include/boost/wave/util/macro_helpers.hpp
jefftrull: [the caller](https://github.com/jefftrull/wave/blob/9121d11e20adb392160f2d4ec8f19a64a901677b/include/boost/wave/util/cpp_macromap.hpp#L1044) reports an error, in this case.

test/testwave/testfiles/t_8_001.cpp
jefftrull: 1. I will add hook verification.
2. I will investigate whether the placemarkers are new, and if not, look at removing them from string reps.

include/boost/wave/util/cpp_iterator.hpp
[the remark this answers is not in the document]
hkaiser: Correction: P0306r4 says, that `__VA_OPT__` shall always occur as part of the token sequence `__VA_OPT__(content)`, so using it without parenthesis the program is ill-formed. 

It is not entirely clear whether that means that only the eventually expanded token sequence should form the `__VA_OPT__(content)` (i.e. whether it is allowed to construct this sequence while expanding macros such that the `__VA_OPT__` may appear without parenthesis and the required sequence is piecewise constructed) or whether any occurrence of `__VA_OPT__` without parenthesis is ill-formed.

include/boost/wave/util/cpp_macromap.hpp
jefftrull: No, and I have [a test](https://github.com/jefftrull/wave/blob/9121d11e20adb392160f2d4ec8f19a64a901677b/test/testwave/testfiles/t_8_007.cpp#L15) for that.

include/boost/wave/util/cpp_macromap_predef.hpp
jefftrull: I'm inclined to leave it out :) but maybe @eldiener would have an opinion on the subject.

include/boost/wave/util/cpp_iterator.hpp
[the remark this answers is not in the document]
hkaiser: IOW, is the following valid code or not:
```c++
#define MAKE_FUNC(x, ...) x __VA_ARGS__
#define F(x, f, ...)  x(f MAKE_FUNC(__VA_OPT__, (,)) __VA_ARGS__))
F(print, "%d", 1)
```

include/boost/wave/util/cpp_macromap.hpp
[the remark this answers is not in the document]
hkaiser: Ok, I'll investigate this further.

include/boost/wave/util/cpp_iterator.hpp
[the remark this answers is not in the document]
jefftrull: Is there a testcase you can think of we could use to see what gcc/clang do?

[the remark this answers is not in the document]
hkaiser: clang reports a `t.cpp:2:47: error: missing '(' following __VA_OPT__` for the example above. I'm inclined to think this is a mistake, however.

include/boost/wave/util/cpp_macromap.hpp
jefftrull: Sure, no problem. Going with:

>            // Before C++20, at least one argument must be supplied for
>            // the variable parameters. So reduce the mandatory
>            // argument count by one if the last parameter is ellipsis:

include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -973,19 +982,29 @@ bool adjacent_stringize = false;
...
             if (IS_EXTCATEGORY((*cit), ExtParameterTokenType)) {
                 BOOST_ASSERT(boost::wave::need_variadics(ctx.get_language()));
                 i = token_id(*cit) - T_EXTPARAMETERBASE;
                 is_ellipsis = true;
             }
             else
+#if BOOST_WAVE_SUPPORT_VA_OPT != 0
+
+            if (IS_EXTCATEGORY((*cit), OptParameterTokenType)) {
+                BOOST_ASSERT(boost::wave::need_va_opt(ctx.get_language()));
```
hkaiser: I think this should actually be an error. Otherwise applications will silently fail if they configure the library for C++11 but encounter a `__VA_OPT__` (I have seen this assertion fire in the `wave --c++11 ...`, which is unexpected and obscure).

tool/cpp.cpp
jefftrull: :man_facepalming:

jefftrull: I did that in *two* places

include/boost/wave/util/cpp_macromap.hpp
jefftrull: I am following the pattern from [this line](https://github.com/jefftrull/wave/blob/60ebbd80e35cd8a7fffc71870f7be22e3a0ad4ce/include/boost/wave/util/cpp_macromap.hpp#L988); is `__VA_OPT__` a different situation in this way?

include/boost/wave/util/cpp_macromap.hpp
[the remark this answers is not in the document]
hkaiser: How about writing it as:
```c++
                    if ((i == arguments.size()) ||             // no variadic argument
                        impl::is_whitespace_only(arguments[i])) {   // no visible tokens
                        // no args; insert placemarker if one is not already present
                        expanded.push_back(
                            typename ContainerT::value_type(T_PLACEMARKER, "\xA7", pos));
                    }
                    else if (!impl::is_blank_only(arguments[i])) {
                        // …
                    }
```

include/boost/wave/util/cpp_iterator.hpp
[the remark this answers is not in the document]
jefftrull: I feel like
> shall always occur as part of the token sequence __VA_OPT__(content)

is pretty specific, and implies the macro `F` is ill-formed. gcc also rejects it...

If Wave were to support this, what changes should be made in the checking code?

include/boost/wave/util/cpp_macromap.hpp
jefftrull: I'll give that a try. Do you think this line from the proposal changes anything?

> Otherwise, the replacement consists of the results of the expansion of content as the replacement list of the current function-like macro before rescanning and further replacement

hkaiser: Hmmm, I think you're right. `__VA_OPT__` is to be treated as a function like macro... Please disregard this comment.

include/boost/wave/util/cpp_macromap.hpp
jefftrull: I will add a test case and whatever code is required to properly check

include/boost/wave/util/cpp_iterator.hpp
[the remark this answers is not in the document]
hkaiser: Ok, let's raise an error then. It's probably impossible to detect whether `__VA_OPT__` is 'eventually' composed to be followed by parenthesis...

include/boost/wave/util/cpp_macromap_utils.hpp
hkaiser: I think this is resolved as discussions above. Thanks!

include/boost/wave/util/cpp_macromap.hpp
hkaiser: Right, it's probably wrong for `__VA_ARGS__` above as well. :/

include/boost/wave/util/cpp_macromap.hpp
jefftrull: OK so I replaced the call to `expand_replacement_list` with this simple copy (which seems to be what `replace_ellipsis` does):
```c++
std::copy(cstart, cit, std::inserter(expanded, expanded.end()));
```
and I get some unit test failures, all of which seem to be that `__VA_ARGS__` is not replaced in the output.

include/boost/wave/util/cpp_iterator.hpp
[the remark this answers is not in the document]
jefftrull: I'll add a unit test, and any code necessary to ensure an error fires.

include/boost/wave/util/cpp_macromap.hpp
jefftrull: I'll add unit tests and modify the checking code.

include/boost/wave/util/cpp_iterator.hpp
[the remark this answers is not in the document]
eldiener: > 
> 
> Correction: P0306r4 says, that `__VA_OPT__` shall always occur as part of the token sequence `__VA_OPT__(content)`, so using it without parenthesis the program is ill-formed.
> 
> It is not entirely clear whether that means that only the eventually expanded token sequence should form the `__VA_OPT__(content)` (i.e. whether it is allowed to construct this sequence while expanding macros such that the `__VA_OPT__` may appear without parenthesis and the required sequence is piecewise constructed) or whether any occurrence of `__VA_OPT__` without parenthesis is ill-formed.

I tnink that __VA_OPT__ not followed by '( optional_content )' is ill-formed. I do not think you can have a __VA_OPT__ followed by something , which expands to '( optional_content )', to make it acceptable. Of course in the form of '__VA_OPT__ ( optional_content )' the optional_content may eventually expand into other tokens but that only occurs if  '__VA_OPT__ ( optional_content )' is resolved to 'optional_content' when the macro output gets rescanned for replacement. Of course I could be wrong but this is my reading of the C++20 standard as it now exists in N4842.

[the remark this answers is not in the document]
jefftrull: Added the code and unit test to the PR

test/testwave/testfiles/t_8_001.cpp
jefftrull: It looks like on develop branch `t_1_032.cpp`, which I haven't touched, also shows placemarkers in the output, so I think this is not new. I will investigate scrubbing them.

hkaiser: Ok, I wasn't even aware of this anymore... In that case, sorry for the noise. It might be even correct to leave them in to ensure proper functioning (for instance to make sure that we don't see two of them in a row, etc.)

include/boost/wave/util/cpp_macromap_predef.hpp
jefftrull: If there are no objections, can we leave this out for now until someone has an application?

include/boost/wave/util/cpp_macromap.hpp
jefftrull: I have an alternative proposal:
Let's keep the assert. "This should never happen" is actually a sensible idea here, because `T_EXTRAPARAMETER`/`T_OPTPARAMETER` are marked by a previous parsing step.
Instead, maybe we can add a check for the language [at that prior point](https://github.com/jefftrull/wave/blob/f6f71ddb8cde9b99dd36db79beb75629e4c779a1/include/boost/wave/util/macro_definition.hpp#L111) - a context is available for passing into `replace_parameters()`...

What do you think?

hkaiser: Sounds good to me! Thanks!

jefftrull: OK so I'm back again :)
I can reproduce this issue with `__VA_OPT__` but not `__VA_ARGS__`. After further investigation I think that `__VA_ARGS__`, and thus the first location, is protected against this assertion firing because:

1. `--c++11` implies `need_variadics()`
2. without `--c++11` (or `--c++20`) defining a macro with ellipsis parameters fails with `ill formed #define directive`
3. without ellipsis no tokens get marked as `EXTPARAMETER`/`OPTPARAMETER`

so I think [the first assertion](https://github.com/jefftrull/wave/blob/60ebbd80e35cd8a7fffc71870f7be22e3a0ad4ce/include/boost/wave/util/cpp_macromap.hpp#L988) is fine. If you can find a test case that shows otherwise I would be glad to see it.

The second one, on the other hand, needs a language check for the case when we run with `--c++11`. I'll put both checks in anyway for the sake of consistency...

hkaiser: Thanks for your investigation and that analysis. I think your proposed solution is fine.

test/testwave/testfiles/t_8_001.cpp
jefftrull: I've reviewed the hook output and concluded it's correct, so I've updated the testcase accordingly.

PR #78, opened 2020-03-08
Repair lexertl lexer and add to test suite

### LINKED ISSUE #77: segmentation fault in list_includes
For some time (I went as far back as 1.41) list_includes gives a segfault on simple input, e.g.:

```c++
#include "foo.h"

int main() {}
```

jefftrull:
I cannot tell why the original code used `dfa_offset` in these pointer
calculations, but from a close review of the dfa state tables it
appears incorrect. Removing it causes list_includes, as well as the
lexertl test, to stop segfaulting and fixes #77.

Perhaps at some point the related code inside Spirit changed?


BenHanson:
Hello!

Here is an example using the latest lexertl (lexertl14):

#include "include/lexertl/generator.hpp"
#include "include/lexertl/stream_shared_iterator.hpp"

int main()
{
    try
    {
        lexertl::rules rules;
        lexertl::state_machine sm;

        // Silly rule set just for demo purposes
        rules.push("a$", 1);
        rules.push("a\r?\n", 2);
        lexertl::generator::build(rules, sm);
        lexertl::table_based_cpp::generate_cpp("lookup", sm, false, std::cout);
    }
    catch (const std::exception &e)
    {
        std::cout << e.what() << '\n';
    }

    return 0;
}

Let me know if you need more info, I remember that lexertl was not updated to the modern interface in Spirit.

Regards,

Ben

jefftrull:
Thanks @BenHanson that plus the remaining hints in the code were enough for me to do the C++03 equivalent, the code for which is still in Spirit...

My last commit updates the PR to include new static tables and comments explaining how to generate them.

PR #81, opened 2020-03-15
Remove linking to newly header-only date_time

jefftrull:
Boost develop branch has an updated date_time that is header-only so
Wave's builds will now fail unless this dependency is removed.


jefftrull:
> Excellent catch!

@pdimov identified the problem right away

PR #85, opened 2020-04-08
Fix more allocator traits

### LINKED ISSUE #84: Cannot build with clang-11 and libc++ in C++20 mode.
This component fails to compile due to not using `std::allocator_traits` with clang-11 and libc++ in C++20 mode.

```
boost/wave/util/flex_string.hpp:547:25: error: no type named 'size_type' in 'std::__1::allocator<char>'
    typedef typename A::size_type size_type;
            ~~~~~~~~~~~~^~~~~~~~~
```

* [build log](https://github.com/boostorg/wave/files/4447651/build-log.txt)
* [config.jam](https://github.com/boostorg/wave/files/4447652/config.jam.txt)

Command used to build boost:

```
b2 --layout=system
   --project-config=/mnt/c/Workspace/vcpkg/buildtrees/boost/x64-linux-dbg/config.jam
   --prefix=/mnt/c/Workspace/vcpkg/packages/boost_x64-linux/debug
   --libdir=/mnt/c/Workspace/vcpkg/packages/boost_x64-linux/debug/lib
   --exec-prefix=/mnt/c/Workspace/vcpkg/packages/boost_x64-linux/debug/bin
   --includedir=/mnt/c/Workspace/vcpkg/packages/boost_x64-linux/debug/../include
   --cmakedir=/mnt/c/Workspace/vcpkg/packages/boost_x64-linux/debug/share/boost
   --build-dir=/mnt/c/Workspace/vcpkg/buildtrees/boost/x64-linux-dbg/build
   --without-graph_parallel --without-mpi --without-python
   --includedir=/mnt/c/Workspace/vcpkg/packages/boost_x64-linux/include
   architecture=x86 address-model=64
   variant=debug threading=multi
   link=static runtime-link=shared
   debug-symbols=on install
```

The `log` component has a similar issue: https://github.com/boostorg/log/issues/114

jefftrull:
If merged this will resolve #84. Thanks to @qis for the assistance.

### DIFF
--- a/include/boost/wave/util/flex_string.hpp
+++ b/include/boost/wave/util/flex_string.hpp
@@ -287,29 +287,31 @@ inline bool operator!=(const mallocator<T>&,
   return false;
 }
 
-template <class Allocator>
-typename Allocator::pointer Reallocate(
-    Allocator& alloc,
-    typename Allocator::pointer p,
-    typename Allocator::size_type oldObjCount,
-    typename Allocator::size_type newObjCount,
-    void*)
-{
-    // @@@ not implemented
-    return NULL;
-}
+#if defined(BOOST_NO_CXX11_ALLOCATOR)
+template<class A>
+struct allocator_traits {
+    typedef typename A::value_type value_type;
+    typedef typename A::pointer pointer;
+    typedef typename A::const_pointer const_pointer;
+    typedef typename A::size_type size_type;
+    typedef typename A::reference reference;
+    typedef typename A::const_reference const_reference;
 
-template <class Allocator>
-typename Allocator::pointer Reallocate(
-    Allocator& alloc,
-    typename Allocator::pointer p,
-    typename Allocator::size_type oldObjCount,
-    typename Allocator::size_type newObjCount,
-    mallocator<void>*)
-{
-    // @@@ not implemented
-    return NULL;
-}
+    static pointer allocate( A& a, size_type n, const_pointer hint )
+    {
+        return a.allocate(n, hint);
+    }
+
+    static void deallocate( A& a, pointer p, size_type n )
+    {
+        a.deallocate(p, n);
+    }
+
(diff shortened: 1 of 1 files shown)


glenfe:
@jefftrull Something like:

```cpp
namespace boost {
namespace wave {
namespace util {

#if !defined(BOOST_NO_CXX11_ALLOCATOR)
using std::allocator_traits;
#else
template<class A>
struct allocator_traits {
    typedef typename A::value_type value_type;
    typedef typename A::pointer pointer;
    // Fill in the rest as needed.
};
#endif

} // util
} // wave
} // boost
```

jefftrull:
Looks like this problem has been solved by a few other libraries and appears to be in the public interface of Container - is it worth simply using that one? https://www.boost.org/doc/libs/1_72_0/doc/html/boost/container/allocator_traits.html

glenfe:
@jefftrull That would be fine too if there are no objections to Boost.Wave taking a dependency on Boost.Container (which it doesn't right now, I believe).

Lastique:
@glenfe @igaztanaga Another option would be to move `allocator_traits` from Boost.Container to Boost.Core. There are other places where we need `allocator_traits` but not the rest of Boost.Container.

jefftrull:
Upon reflection I don't think it makes sense to add a dependency for the benefit of this one string class within one library. Unless `allocator_traits` goes into Core I'll stick with a minimal local version.

jefftrull:
The Travis job has been hung for > 24h so I'm just going to merge it.

glenfe:
@Lastique https://github.com/boostorg/core/commit/c31e23b36294bc3958b04deef62123abce0204b9

Lastique:
@glenfe Thank you Glen.

jefftrull:
Will refactor to use that in 1.74.

PR #86, opened 2020-04-10
Fix more allocator traits (#85)

jefftrull:
Finish the process of replacing allocator member types with allocator traits using an approach suggested by @glenfe.


jefftrull:
Merge on hold pending the release of 1.73-beta

PR #91, opened 2020-05-11
Remove use of deprecated header boost/detail/iterator.hpp

Lastique:
The header is deprecated in favor of `<iterator>`. It generates compiler warnings and will be removed in a future release.

Also, added a missing include for `iterator_value` trait.

### DIFF
--- a/samples/token_statistics/xlex/xpressive_lexer.hpp
+++ b/samples/token_statistics/xlex/xpressive_lexer.hpp
@@ -16,9 +16,9 @@
 #include <string>
 #include <vector>
 #include <utility>
+#include <iterator>
 #include <algorithm>
 
-#include <boost/detail/iterator.hpp>
 #include <boost/xpressive/xpressive.hpp>
 
 namespace boost {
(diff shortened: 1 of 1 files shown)


jefftrull:
Thanks for the PR @Lastique . Will this code work OK with C++03?

Lastique:
Yes, I see no reason why it shouldn't.

jefftrull:
Thank you. Do we need `<iterator>` at all? I don't see us using any iterator traits from `std`. Should we remove the Boost traits entirely and use `std::iterator_traits<Iterator>::value_type` instead?

Lastique:
There is a call to `std::advance` down in the code.

jefftrull:
Ah, OK. What do you think about replacing `boost::iterators::iterator_value` with `std::iterator_traits<>::value_type` and eliminating the Boost header entirely?

Lastique:
Sure, I can do that.

Lastique:
Done.

jefftrull:
@Lastique Travis failed with a Clang issue related to Filesystem - it looks like you were the last to touch the associated file :) Would you be willing to look at [the failure](https://travis-ci.org/github/boostorg/wave/jobs/685844572)?

Lastique:
That failure is unrelated to this PR. Looks like `struct stat` has unexpected layout on Darwin, which makes Boost.Filesystem fail to compile. I'll take care of that in Boost.Filesystem.

Lastique:
You can restart the CI job, it should succeed now.

PR #93, opened 2020-05-26
Simplify allocator use via core/allocator_access

glenfe:


jefftrull:
LGTM too but out of curiosity - what's the motivation for the use of `allocator_XXX_type<A>::type` as opposed to `allocator_traits<A>::XXX_type` as used in the standard?

glenfe:
As we've found in the standards committee, `allocator_traits` as one big collection of traits is mistake. Users specialize it, and then we can't extend it without breaking user specializations. We realize that this (individual traits) is the way we should have had it since the beginning.

PR #101, opened 2020-06-23
Initial set of 1.74 changes

jefftrull:
These have had a good long time to simmer in develop and the `__VA_OPT__` support in particular will be nice to release.

I'm still hoping to have something for `__has_include` prior to tomorrow's feature freeze but just in case it seemed wise to prepare this.


jefftrull:
The conflict seems to have something to do with the allocator fixes, which went into both master and develop. I'll track it down.

jefftrull:
Looks like the cleanest fix for the conflict is to directly merge the missing commit from @glenfe, despite the fact it is already in develop...

PR #102, opened 2020-06-24
Implement __has_include()

jefftrull:
Implement C++17 feature (and vendor extension, previously) __has_include()

Includes feature flags and unit tests.


include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -508,6 +530,53 @@ macromap<ContextT>::is_defined(string_type const &str) const
...
+        IteratorT first = begin;
+        IteratorT last = end;
+        ctx.expand_whole_tokensequence(first, last, filetoks);
+    }
+
+    // extract tokens into string
+    std::string fn(boost::wave::util::impl::as_string(filetoks).c_str());
+
+    // trim leading and trailing blanks
+    boost::algorithm::trim_if(fn, is_space);
```
hkaiser, line 567: What about (c-style) comments? Is it a good idea to do this on the character level? It might be better to do the trimming on the token level.

jefftrull: I'm following the pattern [here](https://github.com/boostorg/wave/blob/bacb94f394943f61517406ad0136093bebdc8158/include/boost/wave/util/cpp_iterator.hpp#L1722), where includes are handled by combining the tokens, then trimming. Could there be comments inside a `__has_include()`?

hkaiser: > I'm following the pattern here, 

Should you use the existing `trim_whitespace()` in this case?

> where includes are handled by combining the tokens, then trimming.

Not sure if comments are removed by the parser in this case. I need to check.

> Could there be comments inside a __has_include()?

Sure, why not, it's equivalent to whitespace after all?

jefftrull: Yeah, so... the weird thing is the existing trim code is in an internal namespace in `cpp_iterator.hpp`, and the macro expansion code is in `cpp_macromap.hpp`. I guess this is a side effect of how `__has_include` crosses the existing boundaries between functionality, combining a little "include processing" and a little "macro processing". I struggled with this for a while. You could say that this approach (a minor wheel reinvention) was the one that I disliked the least :) What would you suggest?

I will check how clang/gcc handle this situation.

hkaiser: Couldn't we move the helper code to some commonly included spot (for instance, here: https://github.com/boostorg/wave/blob/develop/include/boost/wave/util/macro_helpers.hpp)?

jefftrull: What if we move the whitespace trimming into util/macro_helpers.hpp? It has similar functionality.

jefftrull: That sounds like a great idea :) (sorry missed your previous message somehow)

jefftrull: I can confirm that gcc handles comments within `__has_include()` and Wave... does not, yet. Thanks for the observation.

jefftrull: Well, that was an easy fix. Comments were removed by the skipper already - I just needed to fix a check I added.

jefftrull: Any other concerns at the moment? I'll forge ahead with unit tests in the meantime.

include/boost/wave/grammars/cpp_has_include_grammar.hpp
```diff
@@ -0,0 +1,215 @@
...
+            using namespace boost::wave;
+            using namespace boost::wave::util;
+
+            has_include_op
+                =   ch_p(T_IDENTIFIER)      // token contains '__has_include'
+                    >>  (
+                        ch_p(T_LEFTPAREN) >>
+                        (system_include | nonsystem_include | computed_include)
+                        )
+                    >> end_p
```
hkaiser, line 85: What about the right parenthesis? Should this be matched as well?

include/boost/wave/util/cpp_macromap_utils.hpp
```diff
@@ -142,7 +142,7 @@ is_special_macroname (StringT const &name)
     if (name.size() < 7)
         return false;
 
-    if ("defined" == name)
+    if (("defined" == name) || ("__has_include" == name))
```
hkaiser, line 145: Should we check for `__has_include` only if it's detection was enabled (C++17)?

include/boost/wave/grammars/cpp_has_include_grammar.hpp
hkaiser: nvm, I found where the right parenthesis is handled. Sorry for the noise.

test/testwave/testfiles/t_2_027.cpp
```diff
@@ -0,0 +1,13 @@
...
+    Boost.Wave: A Standard compliant C++ preprocessor library
+    http://www.boost.org/
+
+    Copyright (c) 2001-2012 Hartmut Kaiser. Distributed under the Boost
+    Software License, Version 1.0. (See accompanying file
+    LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+=============================================================================*/
+
+// Test error reporting during redefinition of '__has_include'
+
```
hkaiser, line 11: Does this miss a `//O --c++20`?

include/boost/wave/grammars/cpp_has_include_grammar.hpp
jefftrull: I'm glad you asked :) I initially did, but then moved it [outside the parser](https://github.com/jefftrull/wave/blob/6dc21c848009b8ffea5471b23a21030f9cbe67fa/include/boost/wave/util/cpp_macromap.hpp#L1812) because of a tricky problem with Classic's greedy matching.

I want to accept expressions including nested parentheses while excluding the final right parenthesis. When I tried to do it with the Kleene star it would always grab every right paren, including the one that terminates `__has_include()` - and it does not backtrack when the full expression `ch_p(T_LEFTPAREN) >> ... >> ch_p(T_RIGHTPAREN)` then fails. The explanation for this problem [here](https://www.boost.org/doc/libs/1_73_0/libs/spirit/classic/doc/rationale.html#exhaustive_rd) did not provide any solution I could understand.

Is there a cleaner way to solve this?

include/boost/wave/util/cpp_macromap_utils.hpp
jefftrull: Yes :) I am actually writing the feature tests (and conditional compilation stuff) right now.
Also I think this is a C++17 feature - so I'm adding a C++17 mode as well.

test/testwave/testfiles/t_2_027.cpp
jefftrull: It will be `//O --c++17`... probably sometime later today.

jefftrull:
Absolutely. I am also going to test for redefinition, as we do with `__VA_OPT__`.

include/boost/wave/grammars/cpp_has_include_grammar.hpp
hkaiser: If you need to accept inner parenthesis a possible solution would be a recursive rule. Something like:
```
r = ch_p(T_LEFTPAREN) >> r >> ch_p(T_RIGHTPAREN) | *char_p;
```

include/boost/wave/util/cpp_macromap_utils.hpp
hkaiser: Ok, C++17, then. In this case you might need to adapt the tests to ask for this as well.

jefftrull:
I'm satisfied with the code at this point and am just waiting for CI to finish. Will wait for any final feedback.

PR #106, opened 2020-07-21
Track the location tokens were expanded from, for __LINE__ and __FILE__

### LINKED ISSUE #94: Tricky line number tracking
In [a recent Twitter post](https://twitter.com/ericniebler/status/1252660334545866752?s=20), @ericniebler asks
> what should be the token sequence emitted by the preprocessor for this program:
```c++
#define FOO(X) __LINE__ BAR
#define BAR(X) __LINE__ 
FOO(X)
   (Y)
```
> GCC says: "3 3" but clang and msvc say "3 4". Who's right?

Sadly, Wave reports "3 1".

jefftrull:
An (optional) extra position field is added to token_data and set for
any identifier token created from a macro expansion. This information
is used to correctly calculate the filename and line number.

If merged, this will resolve #94.

NOT intended for Boost 1.74.

### DIFF
--- a/include/boost/wave/util/cpp_macromap.hpp
+++ b/include/boost/wave/util/cpp_macromap.hpp
@@ -33,6 +33,7 @@
 
 #include <boost/filesystem/path.hpp>
 #include <boost/lexical_cast.hpp>
+#include <boost/optional.hpp>
 
 #include <boost/wave/util/time_conversion_helper.hpp>
 #include <boost/wave/util/unput_queue_iterator.hpp>
@@ -175,7 +176,8 @@ protected:
         unput_queue_iterator<IteratorT, token_type, ContainerT> &first,
         unput_queue_iterator<IteratorT, token_type, ContainerT> const &last,
         bool& seen_newline, bool expand_operator_defined,
-        bool expand_operator_has_include);
+        bool expand_operator_has_include,
+        boost::optional<position_type> expanding_pos);
 
 //  Collect all arguments supplied to a macro invocation
 #if BOOST_WAVE_USE_DEPRECIATED_PREPROCESSING_HOOKS != 0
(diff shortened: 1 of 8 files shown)


jefftrull:
With these changes we get the same answers as gcc - including on the difficult test case from the standard proposal (which I made into a unit test).

include/boost/wave/util/cpp_macromap.hpp
```diff
@@ -2054,6 +2086,25 @@ macromap<ContextT>::is_valid_concat(string_type new_value,
...
+{
+    typename ContainerT::iterator ex_end = tokens.end();
+    for (typename ContainerT::iterator it = tokens.begin();
+         it != ex_end; ++it) {
+        // expand positions are only used for __LINE__, __FILE__, and macro names
+        if (token_id(*it) == T_IDENTIFIER)
+            it->set_expand_position(pos);
+    }
+}
+
```
hkaiser, line 2107: Will all tokens receive the same position information? Should they receive the corresponding column information as well?

jefftrull: IMO, "no" on both. From my investigation, only `__LINE__` and `__FILE__` need to make use of the "position where expanded" information. For implementation reasons general identifiers do as well, in case they are queued instead of immediately expanded - they must pass their own expand positions for use by any `__LINE__` or `__FILE__` macros in their own expansion. That's one of the problems with `BAR` in the test case that triggered this project - it is a function-like macro whose name token and arguments come from separate macro expansions.

As for the columns, I haven't seen any cases where the actual column number is needed. The `expand_pos` fields do capture it, but we don't use them.

hkaiser: @jefftrull thanks for the explanations, makes sense. As for columns, they are used for error messages, I believe.

PR #109, opened 2020-08-11
Introduce C++20 tokens

### LINKED ISSUE #107: Support C++20 keywords
When C++20 support is enabled and the mode entered, Wave should recognize C++20 keywords, of which 8 are listed at [cppreference](https://en.cppreference.com/w/cpp/keyword), as their own tokens instead of identifiers.

jefftrull:
This adds C++20 keywords as tokens to all lexers when in C++20 mode. It also adds C++11 tokens to the lexers that previously lacked them, and updates the token test to cover all.

If merged, this will resolve #107


jefftrull:
This is not for 1.74.

PR #112, opened 2020-09-11
Remove use of non-existent Boost.Config macro BOOST_NO_CXX11_HDR_MEMORY

mclow:
Thanks to Jeff Trull for the catch.


jefftrull:
Thanks, Marshall!

PR #113, opened 2020-09-25
Clarify loop logic to satisfy a Clang warning

### LINKED ISSUE #90: preprocess_pragma_output sample may be broken
Clang gives a warning for [this code](https://github.com/boostorg/wave/blob/2e8eeeeb4d30e22919dbfa057f09e470d04d5863/samples/preprocess_pragma_output/preprocess_pragma_output.hpp#L146), saying:

> variable 'it' is incremented both in the loop header and in the loop body [-Wfor-loop-analysis]

which does seem to be the case. It's unclear whether that was by design and should probably be investigated.

jefftrull:
Clang warns that a for loop in a sample increments an iterator in both
the header and the body. This seems to be done on purpose to skip
whitespace. I used std::advance in the header instead, for clarity.

Also cleaned up an unused type in the same file while I was there.

If merged, this will resolve #90

### DIFF
--- a/samples/preprocess_pragma_output/preprocess_pragma_output.hpp
+++ b/samples/preprocess_pragma_output/preprocess_pragma_output.hpp
@@ -115,7 +115,6 @@ public:
         typename Context::token_type const& option, 
         Container const& values, typename Context::token_type const& act_token)
     {
-        typedef typename Context::token_type token_type;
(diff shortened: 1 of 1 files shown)


hkaiser (approved):
LGTM, thanks!

PR #115, opened 2020-10-21
Prepare for 1.75 release

jefftrull:
"Breaking Changes" deadline is today - seems like a good time for this. Support for C++20 tokens could possibly be classified as "breaking" in the sense that they were lexed differently, before.


hkaiser (approved):
LGTM, thanks for working on this!

hkaiser (commented):
There is also `operator<=>()` that might need special handling, but that can be added later.

jefftrull:
Oh man, haven't given any thought to attributes... or spaceship... guess I'll file some more issues.

jefftrull:
Re: attributes, is `__has_cpp_attribute()` what you were referring to? I haven't gotten any further than filing an issue for it.

hkaiser:
> Re: attributes, is `__has_cpp_attribute()` what you were referring to? I haven't gotten any further than filing an issue for it.

I meant `[[some_attribute]]` which are separate special tokens, like `T_ATTRIBUTE` which has a value of `[[some_attribute]]`.

jefftrull:
Ah... FWIW Clang doesn't seem to have a special token for that. Is there a specific need, or can we let the parser handle that?

hkaiser:
> Ah... FWIW Clang doesn't seem to have a special token for that. Is there a specific need, or can we let the parser handle that?

Ok. Could be a convenience only. However it could be handled by the parser as a sequence of `[` `[` `identifier` `]` `]` tokens...

PR #117, opened 2020-10-23
Make indentation consistent with code nesting

jefftrull:
These changes, a subset of what clang-format generates, ensure code and comments are aligned properly for the level of the code. There were a number of places where comments (many) and code (some) were four spaces to the left of where they should be. This will make the code easier to read.

One advantage of separating out these changes is they can be verified to be whitespace-only with a simple diff.


hkaiser (approved):
LGTM, thanks!

Issue #116, opened 2020-10-21
Support spaceship operator


jefftrull:
C++20 introduces the spaceship operator `<=>` for which Clang, at least, uses a single token. Seems wise to emulate them by introducing one ourselves.

jefftrull:
Closed via #118

PR #118, opened 2020-10-25
Add the spaceship <=> operator to the C++20 tokens

jefftrull:
The only tricky bit seemed to be differentiating between this token and `LESSEQUAL` when not in C++20 mode. Looks like our deadline is Wednesday so this can go in with the other tokens.


jefftrull:
If merged this will resolve #116

hkaiser (approved):
LGTM, thanks! Nice work!

PR #122, opened 2020-11-17
Additional documentation

jefftrull:
* Add documentation for three undocumented hooks

Those are found_unknown_directive, emit_line_directive, and locate_include_file

Also tweak formatting, fix typos.

* Add documentation for five samples

* Add links from samples docs to the hooks they use


jefftrull:
This change only affects documentation, and it's within the window for 1.75 to make such changes.

PR #125, opened 2020-12-13
Remove long-deprecated hooks API

jefftrull:
Removal in 1.76 was noted in the docs and warned about since 1.74.


hkaiser (approved):
LGTM, thanks!

PR #128, opened 2021-03-03
Prepare for 1.76 release

jefftrull:
One CI thing and removing deprecated code.

If I have time in the next 24h I might have another thing to add to this release. If so it will come in the form of a PR into develop.


jefftrull:
Both of these changes were previously approved so this is mainly to get CI running

PR #134, opened 2021-11-03
Remove support for C++03 and prior compilers

jefftrull:
This removes only support for C++98/03 as a compiler for this library - Wave continues to support C++98 and C++03 constructs as a preprocessor.

Support for C++98 and C++03 compilers was deprecated in 1.74, with a warning message.

Doing this enables the use of C++11 constructs in the code, which should help maintainability.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -14,12 +14,12 @@ windowsglobalimage="cppalliance/dronevs2019"
 
 def main(ctx):
   return [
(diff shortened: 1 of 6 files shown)


pdimov:
No objections from me in principle, although I would have done that somewhere in the middle of the release cycle. Today is the deadline for the 1.78 beta, so this change is a bit too disruptive and is probably not entirely allowed by the release schedule (the cutoff for major changes was Oct 27, the cutoff for breaking changes Oct 20).

hkaiser (approved):
LGTM, thanks!

jefftrull:
Per discussion with @pdimov and @glenfe on cpplang Slack and here, this has arrived too late and so we'll do it next cycle. I'll leave this open until master reopens and do a big merge then (with Boost version info updated suitably).

PR #136, opened 2021-12-09
Specify C++11 requirement in top-level Jamfile

jefftrull:
Removing C++03 from Wave's CI config isn't enough - the top-level Boost
build will try to build Wave under unsupported standards/compiler versions
without an explicit requirement in the Jamfile.


hkaiser (approved):
LGTM, thanks!

pdimov (approved):

PR #141, opened 2022-01-02
Qualify "newline at EOF not required" with actually being there

### LINKED ISSUE #137: Simple unknown directive => found_unknown_directive is not called, stripped of pound.
* I am using boost wave, to preprocess glsl. 
* glsl uses a directive at the top, called `#version`.
* Unfortunately the resulting output for some reason just strips the `#` symbol.
* Furthermore, the hook I have to print all unknown directives is not called.
* As a workaround, I have found that using a macro replacement works.

My minimal reproducible example and output: 

```c++
// g++-11 boost-wave-reproduction.cxx  --std=c++17 -lboost_wave -lboost_thread -lboost_filesystem
// ./a.out

#include <string>
#include <filesystem>
#include <iostream>

#include "boost/wave.hpp"
#include "boost/wave/cpp_context.hpp"
#include "boost/wave/cpp_exceptions.hpp"
#include "boost/wave/cpplexer/cpp_lex_iterator.hpp"
#include "boost/wave/language_support.hpp"


template<typename Token>
struct ContextPolicy : boost::wave::context_policies::eat_whitespace<Token> {
  typedef boost::wave::context_policies::eat_whitespace<Token> base_type;


  template <typename ContextT, typename TokenT>
  bool found_directive(ContextT const &ctx, 
      TokenT const &directive) {
    std::cout << "found_directive, directive.get_value(): " << directive.get_value() << std::endl;
    return this->base_type::found_directive(ctx, directive);
  }

  template <typename ContextT, typename ContainerT>
  bool found_unknown_directive(ContextT const& ctx,
      ContainerT const& line, ContainerT& pending) {
    std::cout << "found_unknown_directive, line: ";
    for (const auto& token : line) {
      std::cout << token.get_value() << ",";
    }
    std::cout << std::endl;
    return this->base_type::found_unknown_directive(ctx, line, pending);
  }

  template <typename ContextT, typename TokenT>
  TokenT const& generated_token(ContextT const &ctx, 
      TokenT const& token) {
    std::cout << "generated_token()" << std::endl;
    std::cout << " token.is_valid(): " << (token.is_valid() ? "true" : "false") << std::endl;
    boost::wave::token_id token_id = boost::wave::token_id(token);
    std::cout << " token_id: " << token_id << std::endl;

    if (token.is_valid()) {
      std::cout << " boost::wave::get_token_name(token_id): " << boost::wave::get_token_name(token_id) << std::endl;
      std::cout << " token.get_value(): " << token.get_value() << std::endl;
    }
    return this->base_type::generated_token(ctx, token);
  }
};


int main() {

  typedef boost::wave::cpplexer::lex_iterator<
      boost::wave::cpplexer::lex_token<>>
      lex_iterator_type;
  typedef boost::wave::context<
    /*Iterator=*/std::string::const_iterator,
    /*LexIterator=*/lex_iterator_type,
    /*InputPolicy=*/boost::wave::iteration_context_policies::load_file_to_string,
    /*ContextPolicy=*/ContextPolicy<typename lex_iterator_type::token_type>>
      context_type;

  std::filesystem::path source_path = "/path/to/source.cxx";
  std::string source = R"source(
// This gets replaced by "#version" as expected.
TEST
// The pound sign doesn't appear, and found_unknown_directive is not called.
#version
)source";

  context_type ctx(/*first_=*/source.begin(),
                      /*last_=*/source.end(),
                      /*fname=*/source_path.string().c_str());
  ctx.set_language(/*language=*/boost::wave::language_support::support_cpp11);

  assert(ctx.add_macro_definition(
      /*macrostring=*/"TEST=#version"));
  context_type::iterator_type first = ctx.begin();
  context_type::iterator_type last = ctx.end();


  std::ostringstream sstr;
  while (first != last) {
    sstr << (*first).get_value();
    ++first;
  }

  std::string preprocessed_source = sstr.str();

  std::cout << preprocessed_source << std::endl;
  return 0;
}

```

Output:

```
$ ./a.out 
generated_token()
 token.is_valid(): true
 token_id: 402915724
 boost::wave::get_token_name(token_id): POUND
 token.get_value(): #
generated_token()
 token.is_valid(): true
 token_id: 134480252
 boost::wave::get_token_name(token_id): IDENTIFIER
 token.get_value(): version
generated_token()
 token.is_valid(): true
 token_id: 1476395402
 boost::wave::get_token_name(token_id): NEWLINE
 token.get_value(): 

generated_token()
 token.is_valid(): true
 token_id: 134480252
 boost::wave::get_token_name(token_id): IDENTIFIER
 token.get_value(): version
generated_token()
 token.is_valid(): true
 token_id: 1476395402
 boost::wave::get_token_name(token_id): NEWLINE
 token.get_value(): 

generated_token()
 token.is_valid(): true
 token_id: 1610613137
 boost::wave::get_token_name(token_id): EOF
 token.get_value(): 
generated_token()
 token.is_valid(): false
 token_id: 1610613138
#version
version

```

Expected output:

```
...
found_unknown_directive ...
...
#version
#version
```

jefftrull:
In C++11 and later modes Wave by default does not require newline prior to EOF. This works fine but
there is one place where the test for this feature was used where Wave is not necessarily looking at the end of the file: `pp_is_last_on_line`. One observable (and surprising) result is that unknown
directives are not getting flagged - but there may be others. The unit tests did not discover this
problem because they generally do not set c++11 mode.

This PR adds the necessary qualifiers and adds a unit test (the same as an existing one but with
C++11 enabled).

If merged this will resolve #137

### DIFF
--- a/test/testwave/testfiles/t_2_031.cpp
+++ b/test/testwave/testfiles/t_2_031.cpp
@@ -1,0 +1,20 @@
+/*=============================================================================
+    Boost.Wave: A Standard compliant C++ preprocessor library
+    http://www.boost.org/
+
+    Copyright (c) 2001-2012 Hartmut Kaiser. Distributed under the Boost
+    Software License, Version 1.0. (See accompanying file
+    LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+=============================================================================*/
+
+//O --c++11
+
+// test the error reporting for unknown directives in C++11 mode
+// (newline at EOF not required)
+
+//R
+//E t_2_031.cpp(17): error: ill formed preprocessor directive: #this_is_a_unknown_pp_directive with some parameter
+#this_is_a_unknown_pp_directive with some parameter
+
+//H 21: t_2_031.cpp(17): #this_is_a_unknown_pp_directive with some parameter
+//H 18: boost::wave::preprocess_exception
--- a/test/testwave/testfiles/t_2_031.cpp
+++ b/test/testwave/testfiles/t_2_031.cpp
@@ -2,7 +2,8 @@
     Boost.Wave: A Standard compliant C++ preprocessor library
     http://www.boost.org/
 
-    Copyright (c) 2001-2012 Hartmut Kaiser. Distributed under the Boost
+    Copyright (c) 2001-2012 Hartmut Kaiser.
+    Copyright (c) 2022 Jeff Trull. Distributed under the Boost
     Software License, Version 1.0. (See accompanying file
     LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 =============================================================================*/
@@ -13,8 +14,8 @@
 // (newline at EOF not required)
 
 //R
-//E t_2_031.cpp(17): error: ill formed preprocessor directive: #this_is_a_unknown_pp_directive with some parameter
+//E t_2_031.cpp(18): error: ill formed preprocessor directive: #this_is_a_unknown_pp_directive with some parameter
 #this_is_a_unknown_pp_directive with some parameter
 
-//H 21: t_2_031.cpp(17): #this_is_a_unknown_pp_directive with some parameter
+//H 21: t_2_031.cpp(18): #this_is_a_unknown_pp_directive with some parameter
 //H 18: boost::wave::preprocess_exception
--- a/include/boost/wave/util/cpp_iterator.hpp
+++ b/include/boost/wave/util/cpp_iterator.hpp
@@ -966,7 +966,8 @@ namespace impl {
             if (call_hook)
                 util::impl::call_skipped_token_hook(ctx, *it);
         }
-        return need_no_newline_at_end_of_file(ctx.get_language());
+        return need_no_newline_at_end_of_file(ctx.get_language()) &&
+            ((it == end) || (T_EOF == token_id(*it)));
     }
 
     ///////////////////////////////////////////////////////////////////////////
--- a/test/testwave/testfiles/test.cfg
+++ b/test/testwave/testfiles/test.cfg
@@ -83,6 +83,7 @@ t_2_027.cpp
 t_2_028.cpp
 t_2_029.cpp
 t_2_030.cpp
+t_2_031.cpp
 
 #
 # t_3: Predefined macros


jefftrull:
btw @hkaiser Wave also has the language option `single_line` which is documented as:

> If the `support_option_single_line` flag is set, the Wave library will now report an
>  error if the last line of the processed input is not terminated by a new line...

`no_newline_at_end_of_file` is unfortunately not documented in [the same place](https://github.com/boostorg/wave/blob/develop/doc/class_reference_context.html) - what's the difference? should we add an explanation?

jefftrull:
I'm thinking in particular of [this old trac report](https://svn.boost.org/trac10/ticket/6759) which I believe is referencing [this logic](https://github.com/boostorg/wave/blob/f2957045b5668d544525492b37d071c4f4ddbcc2/include/boost/wave/util/cpp_iterator.hpp#L667)

hkaiser:
> btw @hkaiser Wave also has the language option `single_line` which is documented as:
> 
> > If the `support_option_single_line` flag is set, the Wave library will now report an
> > error if the last line of the processed input is not terminated by a new line...
> 
> `no_newline_at_end_of_file` is unfortunately not documented in [the same place](https://github.com/boostorg/wave/blob/develop/doc/class_reference_context.html) - what's the difference? should we add an explanation?

@jefftrull I don't have any recollection of the difference, sorry :/ My best guess would be that the former is applied in single-line mode, the latter in normal mode.

test/testwave/testfiles/t_2_031.cpp
```diff
@@ -0,0 +1,20 @@
+/*=============================================================================
+    Boost.Wave: A Standard compliant C++ preprocessor library
+    http://www.boost.org/
+
+    Copyright (c) 2001-2012 Hartmut Kaiser. Distributed under the Boost
```
hkaiser, line 5: @jefftrull you should use your copyright for new files (at least)

hkaiser (approved):
LGTM, thanks!

jefftrull:
> > btw @hkaiser Wave also has the language option `single_line` which is documented as:
> > > If the `support_option_single_line` flag is set, the Wave library will now report an
> > > error if the last line of the processed input is not terminated by a new line...
> > 
> > 
> > `no_newline_at_end_of_file` is unfortunately not documented in [the same place](https://github.com/boostorg/wave/blob/develop/doc/class_reference_context.html) - what's the difference? should we add an explanation?
> 
> @jefftrull I don't have any recollection of the difference, sorry :/ My best guess would be that the former is applied in single-line mode, the latter in normal mode.

I think I get it now, perhaps... maybe the documentation should say "... the Wave library should _not_ report an error..."? That would explain that TRAC report. Actually I might rewrite it to explain that it does more than that. I'll send another PR.

PR #142, opened 2022-01-02
Clarify the purpose of the single_line language option

jefftrull:
It currently says Wave "will now" report an error for input which is not terminated by a newline. In fact it "will not" do that.

I also made some descriptive edits that made sense to me personally.


jefftrull:
@wuye9036 FYI if this was your old ticket

hkaiser (approved):
LGTM, thanks!

Issue #145, opened 2022-01-26
Sanitizer complains about reset_version()


realazthat:
Constructing a simple context to process simple source code, with sanitizers on, results in the following report:

```
==295477==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffef510ce2c at pc 0x7fe257fd109f bp 0x7ffef510cc80 sp 0x7ffef510c410
WRITE of size 13 at 0x7ffef510ce2c thread T0
    #0 0x7fe257fd109e in __interceptor_vsprintf ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:1687
    #1 0x7fe257fd12ce in __interceptor_sprintf ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:1730
    #2 0x55f1a0c83119 in boost::wave::util::predefined_macros::reset_version() (/mnt/z/.ite-cache/3c45814383cd30ef7a050405df4dd30d/execroot/__main__/bazel-out/k8-dbg/bin/src/ite/main+0x2c6c119)
    #3 0x55f1a0c83fff in boost::wave::util::predefined_macros::predefined_macros() (/mnt/z/.ite-cache/3c45814383cd30ef7a050405df4dd30d/execroot/__main__/bazel-out/k8-dbg/bin/src/ite/main+0x2c6cfff)
[...]

PR #146, opened 2022-01-27
Clean up sprintf usage

jefftrull:
There have been compiler warnings about `sprintf` usage for some time, and now we have a bug report as well (#145). I replaced them mostly with `boost::format` but in one case with `std::to_string` (our second use of a C++11 feature!).

I also did a few cleanups while I was in the code - indentation and one unused header.


hkaiser (approved):
LGTM, thanks!

PR #149, opened 2022-02-09
Use portable filesystem string conversion instead of native()

jefftrull:
VC++ builds fail on this due to the use of wchar_t


jefftrull:
This was broken in a previous PR but I didn't check MSVC CI results.

hkaiser (approved):
LGTM, thanks!

PR #158, opened 2022-03-09
Fix up C++03 deprecation

jefftrull:
- Add C++11 feature requirements to test and samples subprojects
- Add `constexpr` to the features tested for in the deprecation check
- Add the C++11 requirement to the documentation metadata
- Add one C++03 build to the Appveyor list. It should pass without running (serves as a test on the above).


jefftrull:
This is kind of a fix for the 1.79 release. I intend to merge it tonight (the deadline) if there are no objections

Issue #159, opened 2022-03-10
Warnings about freeing non-heap-allocated object


Lastique:
On Kubuntu 21.10, gcc 11.2 produces warnings like these when building Boost.Wave 1.79.0 beta1 RC1 (presumably, revision f861b1a305a5dd70286ff84255ad78b1fd46975c):

```
In file included from /usr/include/x86_64-linux-gnu/c++/11/bits/c++allocator.h:33,
                 from /usr/include/c++/11/bits/allocator.h:46,
                 from /usr/include/c++/11/string:41,
                 from libs/wave/src/token_ids.cpp:18:
In member function ‘__gnu_cxx::new_allocator<char>::deallocate(char*, unsigned long)’,
    inlined from ‘boost::allocator_deallocate<std::allocator<char> >(std::allocator<char>&, boost::allocator_pointer<std::allocator<char> >::type, boost::allocator_size_type<std::allocator<char> >::type)void’ at ./boost/core/allocator_access.hpp:375:17,
    inlined from ‘boost::wave::util::AllocatorStringStorage<char, std::allocator<char> >::Free(void*, unsigned long)’ at ./boost/wave/util/flex_string.hpp:542:36,
    inlined from ‘boost::wave::util::AllocatorStringStorage<char, std::allocator<char> >::~AllocatorStringStorage()’ at ./boost/wave/util/flex_string.hpp:624:17,
    inlined from ‘boost::wave::util::AllocatorStringStorage<char, std::allocator<char> >::reserve(unsigned long)’ at ./boost/wave/util/flex_string.hpp:678:5,
    inlined from ‘boost::wave::util::CowString<boost::wave::util::AllocatorStringStorage<char, std::allocator<char> >, char*>::CowString(char const*, unsigned long, std::allocator<char> const&)’ at ./boost/wave/util/flex_string.hpp:1259:23,
    inlined from ‘boost::wave::util::flex_string<char, std::char_traits<char>, std::allocator<char>, boost::wave::util::CowString<boost::wave::util::AllocatorStringStorage<char, std::allocator<char> >, char*> >::flex_string(char const*, std::allocator<char> const&)’ at ./boost/wave/util/flex_string.hpp:1481:43,
    inlined from ‘boost::wave::get_token_name(boost::wave::token_id)’ at libs/wave/src/token_ids.cpp:243:64:
[...]

PR #163, opened 2022-06-09
Fix warnings about freeing non-heap-allocated object

Lastique:
This PR makes 3 commits:

- Convert all checks for whether `pData_` points to `emptyString_` to checks for zero `capacity()`. This is a safer check, which should work across shared library boundaries.
- Added compile-time asserts that `pData_` never points to `emptyString_` when `capacity()` is zero. This fixes https://github.com/boostorg/wave/issues/159.
- Removes `AllocatorStringStorage::Realloc`. This function was not used and potentially incorrect, as it would call `Free` on the input buffer, even if it was `emptyString_`.

I did not use `#pragma` to silence the warnings because it seems gcc 11.2 doesn't support disabling this warning (the warnings are still emitted).

### DIFF
--- a/include/boost/wave/util/flex_string.hpp
+++ b/include/boost/wave/util/flex_string.hpp
@@ -544,14 +544,6 @@ class AllocatorStringStorage : public A
             static_cast<const char*>(p));
     }
 
-    void* Realloc(void* p, size_type oldSz, size_type newSz)
-    {
-        void* r = Alloc(newSz);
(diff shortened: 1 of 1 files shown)


jefftrull (approved):
LGTM and will merge in 24h if @hkaiser has no objections. Would be good to get some CI feedback.

hkaiser (approved):
LGTM, thanks!

Issue #160, opened 2022-04-12
one test fails with error C2660 and error C2440 under msvc /permissive- mode


Zhaojun-Liu:
**Issue description:**
One test fails after runtest in MSVC. Could you please take a look?

**Reproduce steps:**
1. git clone ​https://github.com/boostorg/boost.git F:\gitP\boostorg\boost
2. git -C "F:\gitP\boostorg\boost" submodule update --init --recursive
3. open a VS 2019 x86 command prompt and go to F:\gitP\boostorg\boost
4. set _CL_=/permissive-
5. .\b2 -j16 variant=release --build-dir=..\out\x86rel libs\wave\test

[test_95.log](https://github.com/boostorg/wave/files/8471946/test_95.log)

**ErrorMessage:**
test_lexertl_lexer.cpp
F:\gitP\boostorg\boost\libs/wave/samples/list_includes/lexertl/lexertl_lexer.hpp(432): error C2660: 'boost::wave::operator |': function does not take 1 arguments
F:\gitP\boostorg\boost\boost/wave/token_ids.hpp(341): note: see declaration of 'boost::wave::operator |'
F:\gitP\boostorg\boost\libs/wave/samples/list_includes/lexertl/lexertl_lexer.hpp(430): note: while compiling class template static data member 'const boost::wave::cpplexer::lexertl::lexer::lexertl<boost::wave::util::position_iterator<Iterator,Position>,Position>::lexer_data boost::wave::cpplexer::lexertl::lexer::lexertl<boost::wave::util::position_iterator<Iterator,Position>,Position>::init_data_cpp0x[15]'
        with
        [
            Iterator=char *,
            Position=boost::wave::util::file_position<boost::wave::util::flex_string<char,std::char_traits<char>,std::allocator<char>,boost::wave::util::CowString<boost::wave::util::AllocatorStringStorage<char,std::allocator<char>>,char *>>>
        ]
F:\gitP\boostorg\boost\libs/wave/samples/list_includes/lexertl/lexertl_lexer.hpp(825): note: while compiling class template static data member 'boost::wave::cpplexer::lexertl::lexer::lexertl<boost::wave::util::position_iterator<Iterator,Position>,Position> boost::wave::cpplexer::lexertl::lexertl_functor<Iterator,Position>::lexer_'
        with
        [
            Iterator=char *,
[...]

PR #167, opened 2022-06-22
Use constexpr instead of macros for tokenid constants in samples

jefftrull:
Under Visual Studio 2019 (and possibly earlier), building with `/permissive-` causes three samples to fail to compile. Adding the
`/Zc:twoPhase-` option makes them compile successfully again. See bug #160 for more details.

It's unclear what the source of this issue is, but using `constexpr` variables instead of a macro is better practice anyway, and fixes the errors.


hkaiser (approved):
This LGTM, thanks. Shouldn't we however change all of the `token_id` definitions (at least for consistency's sake)?

hkaiser:
> Shouldn't we however change all of the `token_id` definitions (at least for consistency's sake)?

Nvm, I take the last part back.

PR #170, opened 2022-06-24
Fix line directives after endif following else

### LINKED ISSUE #24: Line numbers wrong after conditional section
I have a small problem and I don't know if this can be easily fixed in wave.

I'm using wave to preprocess a file. It correctly handles macro. 
For instance , output of the following code : 

```
#define DEBUG

MyClass::Test()
{
#ifdef DEBUG
	print("Test");
#endif
    this.Test();
}
 
```
is  (B) : 

```
#line 3 "C:\.....\myfile.c"
MyClass::Test()
{
	print("Test");
    this.Test();
}
 
```
Then I apply a home-made linter to the preprocessed file (B) . However, I'm struggling to identify the initial line number from the preprocessed file. 

Is there a way to get more information in the preprocessed file, something similar to : 

```
#line 3 "...myfile.c"
MyClass::Test()
{
#line 6 "...myfile.c"
	print("Test");
#line 8 "...myfile.c"
    this.Test();
}
```
(or any other methods that permits to find the initial line number when macro is used .
Don't hesitate to tell me if I'm not clear enough.

Many thanks.

jefftrull:
The initial part of a conditional block (the if/ifdef/ifndef) takes the "slow" path through the code in cpp_iterator.hpp using the C++ expression grammar, while else and endif directives take the "fast" path. The slow path sets must_emit_line_directive, and it will be cleared after either else or endif is processed. Unfortunately when an else precedes the endif, must_emit_line_directive is cleared and is never set, as only the slow path sets it. The result is proper line directives are issued for the else clause, but not for the subsequent endif.

This change adds setting of must_emit_line_directive in the "fast" path for endif so line directives are considered after an intervening else. It also adds extensive unit tests for line directives.

If merged, this will resolve #24

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -19,6 +19,7 @@ TODO (known issues):
 CHANGELOG
 
 Boost V1.80:
+  - Fixed #24: Line numbers wrong after conditional section
   - Fixed #160: one test fails with error C2660 and error C2440 under msvc /permissive- mode
   - Fixed #161: BOOST_WAVE_THROW_NAME_CTX does not result in a context callback and always throws
   - Fixed #162: When Boost headers are included in the preprocessed translation unit, Wave fails in boost/integer.hpp(99)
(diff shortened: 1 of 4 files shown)


hkaiser (approved):
LGTM, thanks! Excellent analysis!

PR #173, opened 2022-09-05
Use `boost::allocator_max_size` for `SmallStringOpt` to obtain allocator's max size

frederick-vs-ja:
Since direct call to `a.max_size()` may fail to compile since C++11.

I found the fix was missing for `SmallStringOpt` when building a project at work in C++20 mode (where `std::allocator`'s `max_size` is removed).

Related to #85, #86, #93.

### DIFF
--- a/include/boost/wave/util/flex_string.hpp
+++ b/include/boost/wave/util/flex_string.hpp
@@ -995,7 +995,7 @@ public:
     }
 
     size_type max_size() const
-    { return get_allocator().max_size(); }
+    { return boost::allocator_max_size(get_allocator()); }
 
     size_type capacity() const
     { return Small() ? maxSmallString : GetStorage().capacity(); }


jefftrull:
Wave requires C++11 since 1.79 - maybe we can just use `std::allocator_traits<>` ?

frederick-vs-ja:
> Wave requires C++11 since 1.79 - maybe we can just use `std::allocator_traits<>` ?

I think this is OK. But doing so would partially revert the changes in #93, and might violate @glenfe's intent.

jefftrull:
Yes, I think you are right. There is little practical benefit to switching to `std::allocator_traits` and a bit of work. Making this update maintains consistency with the existing code, as well.

hkaiser:
@jefftrull Thanks a lot!

Issue #182, opened 2022-11-17
cpp_re.hpp has a memmove that passes 0/null as arguments => Sanitizer complains.


realazthat:
### Happens here: https://github.com/boostorg/wave/blob/99ccc6e3cbd3fa347dc23f10518aa7d33d31573e/include/boost/wave/cpplexer/re2clex/cpp_re.hpp#L205

My guess is that sometimes memmove is a noop (length is zero) but nulls are passed in, and sanitizer doesn't like this. If so the fix would be to wrap this in a condition.

### Address sanitizer complains:

![image](https://user-images.githubusercontent.com/141522/202524644-4b1f62bf-ea04-4488-a957-05ca879ae0a5.png)

### Values shown in debugger:

![image](https://user-images.githubusercontent.com/141522/202524737-cd6a45b5-1b3d-4ad4-8299-660513243149.png)

jefftrull:
This is UB and definitely requires fixing. I probably also need to make a testcase - is there anything special about the situation? If you have a small repro I'd love to have it.

jefftrull:
That's fantastic, actually! Thank you.

jefftrull:
This will be fixed in 1.82, to be released soon.

PR #185, opened 2023-03-27
Fix UB in memmove

AmesingFlank:
boost::wave sometimes calls memmove with a nullptr destination and a 0 byte length. This is undefined behavior. This diff adds a check so that memmove is not called if the length is 0


hkaiser (approved):
LGTM, thanks!

AmesingFlank:
> Thanks! Do you have a testcase by any chance? Then we could add it to the regression tests.
> 
> Also this fixes bug #182 :)

I don't have a repro case that I can post here, unfortunately. Can we use the code mentioned in #182 as the regression test?

jefftrull:
Thanks for your contribution, which will be in Boost 1.82 (releasing soon).

PR #187, opened 2023-03-29
Fix memmove UB for 1.82 release

jefftrull:


jefftrull:
I got the OK to merge this in the bugfix window for 1.82, which is now. Just waiting for CI to complete.

PR #190, opened 2023-10-23
Make updates for CMake version, dependencies, and C++ standard

### LINKED ISSUE #189: C++11 requirements for CMake need tweaking
Per @pdimov 's [post](https://lists.boost.org/Archives/boost/2023/10/255042.php) on boost-devel, there are a few requirements for CMake that I didn't update correctly when we began requiring C++11. In particular:

- `cmake_minimum_required` needs to be at least 3.8
- there is no `target_compile_features` specifying C++11
- the dependency on `Boost.format` is not reflected

Generally speaking, the CMake build flow is a parallel one that needs to be maintained as we change the bjam stuff.

jefftrull:
These issues were noted by @pdimov and the fix was to rerun boostdep --cmake (with a couple of manual changes for src/cpplexer files).

If merged this will resolve #189


pdimov:
It's OK, but why reorder the source files? It just makes the diff more complicated for no good reason.

jefftrull:
My thinking is that boostdep is now producing them in a different order, so by taking the reordering now we will have shorter diffs in the future.

I guess I could just always reorder them to match going forward instead. Hm... What do you think?

pdimov:
The order in which the files are enumerated might be OS-dependent; on Windows, `boostdep` produces the same sorted list, but you are probably on Linux or Mac. Well, not that big of a deal, I suppose.

pdimov (approved):

Issue #188, opened 2023-07-13
Segmentation fault when "#pragma\n" is encountered


mfep:
### System info
- Boost version: 1.82.0
- Compiler: g++ 13.0.1-x86_64-redhat-linux
- Both in optimized and unoptimized build

### Actual behaviour
Preprocessing C++ source code with a `#pragma` statement immediately followed by a newline (`\n`) using Wave crashes the program with a segmentation fault.

### Expected behaviour
The preprocessor lets the `#pragma` statement through to the output as unrecognized. Or throw a recoverable error.

### Minimal reproducer
Minimal reproducer based on the "Quick start" section in the documentation:

```c++
#include <boost/wave/cpplexer/cpp_lex_iterator.hpp>
#include <boost/wave.hpp>

#include <iostream>
#include <string>

int main()
{
    std::string input =
        R"(int main()
{
#pragma
    return 0;
}
)";

    typedef boost::wave::cpplexer::lex_iterator<
        boost::wave::cpplexer::lex_token<>>
        lex_iterator_type;
    typedef boost::wave::context<
        std::string::iterator, lex_iterator_type>
        context_type;

    context_type ctx(input.begin(), input.end());

    context_type::iterator_type first = ctx.begin();
    context_type::iterator_type last = ctx.end();

    try
    {
        while (first != last)
        {
            std::cout << (*first).get_value();
            ++first;
        }
    }
    catch (const boost::wave::preprocess_exception &ex)
    {
        std::cout << "ERROR: " << ex.description() << std::endl;
        return -1;
    }
}
```

hkaiser:
Thanks for reporting, I'll try to have a look.

jefftrull:
Thanks for the very clear bug report.

It looks like the code that implements pragmas assumes there is always at least one token (the space) after `#pragma` and when it is not present we can dereference invalid memory. You can see this [here](https://github.com/boostorg/wave/blob/61abf8b6b98d8937515a4eddf6d0d7274921dcda/include/boost/wave/util/cpp_iterator.hpp#L2432) where we unconditionally increment the iterator representing the start of the token sequence and then dereference it.

It looks like there's a simple fix: testing for an empty token sequence and returning early. It may not be the best fix, but it should be a usable workaround. Add this at the start of `on_pragma`:

```
    if (begin == end)
        return true;
```

hkaiser:
I think another possible fix would be to change this:
[...]

PR #191, opened 2023-10-24
Support a single #pragma without body

jefftrull:
This approach differs from the initial plan - which just returned when an empty pragma was detected - because we still need to fire pragma hooks.

### DIFF
--- a/include/boost/wave/util/cpp_iterator.hpp
+++ b/include/boost/wave/util/cpp_iterator.hpp
@@ -2427,10 +2427,9 @@ pp_iterator_functor<ContextT>::on_pragma(
     const_tree_iterator_t last = make_ref_transform_iterator(end, get_value);
 
     expanded.push_back(result_type(T_PP_PRAGMA, "#pragma", act_token.get_position()));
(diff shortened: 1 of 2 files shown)


jefftrull:
If merged this will resolve #188

hkaiser (approved):
LGTM, thanks!

include/boost/wave/util/cpp_iterator.hpp
```diff
@@ -2456,7 +2455,6 @@ pp_iterator_functor<ContextT>::on_pragma(
 #endif
         }
     }
-    expanded.push_back(result_type(T_NEWLINE, "\n", act_token.get_position()));
```
rlenhardt, line 2459: Removing the final new line will cause emitted pragmas to miss the final new line and be misintepreted by following compile steps. For example
```
#pragma foo
#define X hello
X
```
will currently get preprocessed to
```
#pragma foo#define X hello
X
```
but should be actually
```
#pragma foo
hello
```
Shall i add an issue with more details?

PR #192, opened 2023-10-24
Replace deprecated boost::filesystem methods with their new names

jefftrull:
This will be for Boost 1.84 forward. The new replacements were available previously, so for some versions before 1.84 we'll still see deprecation warnings if users mix versions of Boost and Wave.

In the meantime the builds will get quieter.


hkaiser (approved):
LGTM, thanks!

PR #199, opened 2024-01-17
Upgrade stale CI config

jefftrull:
Two of the GitHub Actions runners were no longer valid, so clang Linux and MacOS never ran. Also there are newer versions of several compilers, as well as the Standard.


jefftrull:
I'm going to squash merge this to get rid of all the experimental commits.

PR #204, opened 2024-01-28
Use correct string constructor for non-null-terminated range

jefftrull:
There is no guarantee (indeed, it is unlikely) that data within the scanner buffer will be null terminated and so the constructor that accepts a pointer and a count is appropriate.

This is one of the issues revealed by the `BOOST_WAVE_BSIZE` variation experiment


jefftrull:
t_5_001 reveals this bug with BOOST_WAVE_BSIZE=43 and the address sanitizer turned on

hkaiser (approved):
Excellent catch! LGTM! Thanks.

PR #205, opened 2024-01-29
Fixes for escaped newline handling

jefftrull:
Experiments with `BOOST_WAVE_BSIZE` revealed that we were missing some backslash+newline cases, and in others, testing invalid memory locations for the presence of a newline.

These fixes repair the specific issues found but a more general solution is probably in order (and will be particularly needed given the more lenient treatment of newline escapes in C++23)

### DIFF
--- a/include/boost/wave/cpplexer/re2clex/cpp_re.hpp
+++ b/include/boost/wave/cpplexer/re2clex/cpp_re.hpp
@@ -238,7 +238,7 @@ uchar *fill(Scanner<Iterator> *s, uchar *cursor)
         for (p = std::max(s->lim - 3, s->cur); p < s->lim + cnt - 2; ++p)
         {
             int len = 0;
-            if (is_backslash(p, s->lim + cnt, len))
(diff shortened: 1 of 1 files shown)


include/boost/wave/cpplexer/re2clex/cpp_re.hpp
```diff
@@ -235,10 +235,10 @@ uchar *fill(Scanner<Iterator> *s, uchar *cursor)
         /* backslash-newline erasing time */
 
         /* first scan for backslash-newline and erase them */
-        for (p = s->lim; p < s->lim + cnt - 2; ++p)
+        for (p = std::max(s->lim - 3, s->cur); p < s->lim + cnt - 2; ++p)
```
hkaiser, line 238:
```suggestion
        for (p = (std::max)(s->lim - 3, s->cur); p < s->lim + cnt - 2; ++p)
```

hkaiser (commented):
Could you add some somment to the code you changed for us to be able to remember what's happening?

jefftrull:
Thanks for the feedback; LMK what you think.

hkaiser (approved):
LGTM, thanks!

PR #207, opened 2024-02-02
Fix checks for signed overflow

### LINKED ISSUE #197: Test for signed overflow occurs after UB invoked
I noticed a warning in CI today for [two](https://github.com/boostorg/wave/blob/abe718abf4d9bdfceeaa0ec2f3115bcfb16bd695/include/boost/wave/grammars/cpp_expression_value.hpp#L189) [lines](https://github.com/boostorg/wave/blob/abe718abf4d9bdfceeaa0ec2f3115bcfb16bd695/include/boost/wave/grammars/cpp_expression_value.hpp#L256) in `cpp_expression_value.hpp`:

`warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false`

It looks like we perform the calculation, and _then_ check to see if an overflow resulted. The compiler may discard the check (or worse).

jefftrull:
When verifying that code being preprocessed does not invoke undefined behavior, Wave does it itself.
This change performs tests on the operands in advance, instead.

Tests `t_6_15`, `t_6_17`, and `t_6_18` (though not `t_6_17`, for some reason) signal errors in the absence of this fix with `-fsanitize=signed-integer-overflow` configured, but it does not cause the tests to fail, because these tests are expected to exit with an error anyway.

If merged, this will fix #197

### DIFF
--- a/include/boost/wave/grammars/cpp_expression_value.hpp
+++ b/include/boost/wave/grammars/cpp_expression_value.hpp
@@ -15,6 +15,8 @@
 #include <iostream>
 #endif // defined(BOOST_SPIRIT_DEBUG)
 
+#include <limits>
+
 #include <boost/wave/wave_config.hpp>
 #include <boost/wave/grammars/cpp_value_error.hpp> // value_error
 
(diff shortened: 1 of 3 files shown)


jefftrull:
While reviewing this old PR I realized that we don't handle `INT_MIN / -1` properly so I've added that.

PR #214, opened 2024-07-20
Add support for modular build structure.

grafikrobot:
This is part of the effort to make the Boost libraries "modular" for build and consumption. See https://lists.boost.org/Archives/boost/2024/01/255704.php and https://github.com/grafikrobot/boost-b2-modular/blob/b2-modular/README.adoc for more information.

This PR depends on the following other PRs being merged to both develop and master branches of the respective repos:

- https://github.com/boostorg/boost/pull/854

This PR will be changed to ready for review, i.e. not draft, when the above are merged. Do not merge this one until that time.

Issue #164, opened 2022-06-16
Support "named" variadic macros


jefftrull:
This piece of code appears in recent Ubuntu distributions, in `/usr/include/linux/stddef.h`:

```c++
#define __struct_group(TAG, NAME, ATTRS, MEMBERS...) \
        union { \
                struct { MEMBERS } ATTRS; \
                struct TAG { MEMBERS } ATTRS NAME; \
        }
```

The [gcc manual](https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html) notes that names other than `__VA_ARGS__` - in this case, `MEMBERS` - may be used to refer to the variadic arguments "as an extension". Given the ubiquity of this header it may be useful to add support for this extension.

hkaiser:
Isn't this dangerous/ambiguous? Wasn't the comma right before the `...` optional? I don't remember the details...

jefftrull:
It seems unambiguous to me. If there is some identifier token prior to the `...` we use that instead of `__VA_ARGS__`.

Can you give an example?

hkaiser:
> It seems unambiguous to me. If there is some identifier token prior to the `...` we use that instead of `__VA_ARGS__`.
> 
> Can you give an example?

Most likely I'm misremembering things. Please feel free to implement this gcc extension if you feel it is valuable. However, as wave was always meant to be strictly conforming, I'd suggest enabling this feature only based on an explicit option the user has to provide. Wave could even mention this new option in the error message if it discovers the construct, but the option was not enabled.

jwnhy:
Is this still ongoing?

I need this feature to compile kernel code and I'd like to help.
But I am quite new to Wave too...
Could anyone provide rough guidance about which file I should learn and modify to add this feature?

jefftrull:
Like many open source libraries, Wave progresses at a rate determined by the energy level and availability of its maintainers :) I think your contribution would be very welcome, particularly given the use of this syntax in the Linux kernel (a popular target of analysis)! I'd be happy to help you get involved.

Given that this is a non-standard extension we would need to create a feature flag for it. We support certain MSVC extensions with a macro flag `BOOST_WAVE_SUPPORT_MS_EXTENSIONS` at compile time; another model would be to connect into the language support feature flags available at runtime in [this file](https://github.com/boostorg/wave/blob/develop/include/boost/wave/language_support.hpp). The latter might be more appropriate given that the lexer itself (probably) does not need to be modified.

[...]

PR #218, opened 2024-09-08
Add support for GNU's named variadics for macros.

jwnhy:
As mentioned in #164, named_variadics like `#define(x...) x` is widely used in projects like the Linux kernel.

This PR adds support for this feature.

- Options and exceptions.
This PR adds a `BOOST_WAVE_SUPPORT_GNU_NAMED_VARIADICS_PLACEMARKERS` and 
`support_option_named_variadics = 0x100000`.

When the macro is **enabled**, but the runtime option is **not**, it throws `bad_define_statement_named_va_args`, to remind the user to enable the option.
When the macro is **disabled**, it just reports `ill_formed.`

> I haven't figured out a way to inform the user that he/she needs to enable the macro and recompile for this feature to work...

- Lexer part.
I modified the lexer so it accept construct like `(ch_p(T_IDENTIFIER) >> *ppsp >> ch_p(T_ELLIPSIS))`, then, I used `token_node_d` to *compress* all these node into one, and `access_node_d` to modify the node information to `T_GNU_NAMED_VARIADIC`.

This helps deal with later replacement and expansion.
> Keeping all the nodes makes `(x, y...)` and `(x, y, ...)` indistinguishable in the later phases, as the `T_COMMA` does not generate any node.
> Unless we re-visit the original token seq in the later process,which leads to an egg-chicken problem.

- Expansion part.
I re-use most of the original codes.

- Testing
I have run some basic examples to ensure this feature works as expected.
The original test suite also passed, except for the one failure `t_5_007.cpp` that was introduced in #214, which I guess it's my local `libc` issue.

Let me know if there is any issue.

### DIFF
--- a/include/boost/wave/cpp_exceptions.hpp
+++ b/include/boost/wave/cpp_exceptions.hpp
@@ -267,8 +267,7 @@ public:
             "ill formed #define directive",             // bad_define_statement
             "__VA_ARGS__ can only appear in the "
             "expansion of a C99 variadic macro",        // bad_define_statement_va_args
-            "named variadic like x... is disabled, "
-            "please enable_named_variadics() to enable",  // bad_define_statement_named_va_args
+            "named variadic like x... is disabled",     // bad_define_statement_named_va_args
             "__VA_OPT__ can only appear in the "
             "expansion of a C++20 variadic macro",      // bad_define_statement_va_opt
             "__VA_OPT__ must be followed by a left "
--- a/include/boost/wave/cpp_exceptions.hpp
+++ b/include/boost/wave/cpp_exceptions.hpp
@@ -116,6 +116,7 @@ public:
         ill_formed_operator,
         bad_define_statement,
         bad_define_statement_va_args,
+        bad_define_statement_named_va_args,
         bad_define_statement_va_opt,
         bad_define_statement_va_opt_parens,
         bad_define_statement_va_opt_recurse,
@@ -209,6 +210,7 @@ public:
         case preprocess_exception::unbalanced_if_endif:
         case preprocess_exception::bad_define_statement:
         case preprocess_exception::bad_define_statement_va_args:
+        case preprocess_exception::bad_define_statement_named_va_args:
         case preprocess_exception::bad_define_statement_va_opt:
         case preprocess_exception::bad_define_statement_va_opt_parens:
         case preprocess_exception::bad_define_statement_va_opt_recurse:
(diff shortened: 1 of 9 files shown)


jwnhy:
I was worried about the validity checks around this feature.
After some digging, I have the following observation.

- It seems we don't need to modify the code in [here](https://github.com/boostorg/wave/blob/46bd710ada96c642008a3e6bb387f6738376d758/include/boost/wave/util/cpp_iterator.hpp#L1714) as it looks better if we prevent the usage of `__VA_ARGS__` even if the user has enabled the named variadics.
- The duplicate parameters `#define dup(x,x...)` are handled in [here](https://github.com/boostorg/wave/blob/7ce8b53d9dc56a9c0b9a8855f26003e6b174301f/include/boost/wave/util/cpp_macromap.hpp#L389).

So, I guess the validity checking code *probably* does not have to be added for named variadics?

Also, is there a dependency system for the compilation options?

For now, all `NAMED_VARIADICS` is guarded by `VARIADICS`, so if `VARIADICS` is not enabled, `NAMED_VARIADICS` is ignored.
A thing that would happen is that a user might only enable `NAMED_VARIADICS` without `VARIADICS`. This causes a compilation error if the user tries something like `need_named_variadics(ctx.get_language())`

I added a warning for that issue.

jefftrull:
I haven't had an opportunity to review the code yet (currently traveling) but this will definitely need tests.

You can either add to one of the existing variadic macro tests, or create a new one (particularly if different command line options are required, e.g. whatever activates the gcc extensions mode).

AFAIK the best documentation on the various test features are found in the test driver code [here](https://github.com/boostorg/wave/blob/ef14767aec712d6d342dc598fbb1432236d930c8/test/testwave/testwave_app.cpp#L431). I often start by finding a similar test to the one I need to write under `test/testwave/testfiles`.

jwnhy:
> I haven't had an opportunity to review the code yet (currently traveling) but this will definitely need tests.
> 
> You can either add to one of the existing variadic macro tests, or create a new one (particularly if different command line options are required, e.g. whatever activates the gcc extensions mode).
> 
> AFAIK the best documentation on the various test features are found in the test driver code [here](https://github.com/boostorg/wave/blob/ef14767aec712d6d342dc598fbb1432236d930c8/test/testwave/testwave_app.cpp#L431). I often start by finding a similar test to the one I need to write under `test/testwave/testfiles`.

Sure, thank you for the advice, sorry I am occupied by some other tasks.

Will find a time to add some tests.

Issue #225, opened 2024-09-25
Line directive emitted in middle of multi-line string which contains an ifdef or ifndef


njnobles:
Hi, I'm running into an issue where a line directive is being emitted in the middle of a multi-line string if there's an ifdef or ifndef directive in the middle of the string. The line directive was not emitted in Boost 1.77.0.
I suspect this issue was introduced in this PR: https://github.com/boostorg/wave/pull/140
After reverting that change, I no longer see line directives emitted in the middle of the multi-line string.
Here's an example to illustrate the issue:
**test.cpp**
```
"1
#ifdef A
    2a"
#else
    2b
    3"
#endif
```
The following are using `default_preprocessing_hooks`.
**Boost 1.77.0 preprocessed output:**
```
"1
    2 b
    3"
```
**Boost 1.86.0 preprocessed output:**
```
"1
    #line 5 "t_2_032.cpp"
2 b
    3"
```
Boost 1.86.0 output is identical to Boost 1.77.0 output if I revert https://github.com/boostorg/wave/pull/140

**And for comparison, here's output from Visual Studio's cl.exe preprocessor:**
```
#line 1 "test.cpp"
"1



    2b
    3"
#line 8 "test.cpp"
```

jefftrull:
I'm wondering why the tokenizer doesn't catch that one... Thanks for the report!

njnobles:
To be fair, it is a strange case to run into. But my partners have a knack for abusing preprocessor functionality :)

I've created https://github.com/boostorg/wave/pull/226 which reverts the code change from https://github.com/boostorg/wave/pull/140. I left the new test from 140 in however and, interestingly, it seems to still pass in my local environment.

jefftrull:
Unfortunately changes to test code aren't properly propagated in the build system - try touching [test.cfg](https://github.com/boostorg/wave/blob/develop/test/testwave/testfiles/test.cfg) to rerun.

njnobles:
I did notice that occurring and figured out that trick. This particular test is run as default_hooks.exe. I was able to trigger it to rerun by removing the default_hooks.run and default_hooks.test files. I also ran it directly to make sure it was actually running. Seems like it passed in the PR automation as well.

jefftrull:
I was able to identify the commit that separately (from #140) fixed this problem - it's #170. Now to understand why...

jefftrull:
[...]

PR #140, opened 2022-01-02
Fix line directives for ifdef and ifndef when default hooks are used

jefftrull:
At some point in the past the handling for `#if` and `#ifdef` diverged. The code that handles emitting a line directive when a conditional section is skipped worked for `#if` but not `#ifdef/#ifndef`.

This problem was not observable when the `eat_whitespace_hooks` were used instead of the `default_preprocessing_hooks`, because the former signals skipped newlines through the `may_skip_whitespace` hook, hiding the problem. Furthermore, the majority of Wave tests use the
`eat_whitespace` hooks, so it wasn't visible in testing.

This change restores `#ifdef/#ifndef` to the same section as `#if`, so any future changes to conditional handling will happen uniformly. Also, a test case is added to cover the default hooks and this particular case.

If merged, this will resolve #138

PR #223, opened 2024-09-20
Emit line directives for includes with directive on first line

njnobles:
Attempts to address https://github.com/boostorg/wave/issues/222

Include files that begin with an `#if`/`#ifdef`/`#define` directive now properly emit the line directive that was previously missing. 

Because the test suite is using the `eat_whitespace` context policy, the issue only shows if the directive in the include file occurs on the very first line of the file (hence why the copyright comments have been moved to the end of the test files). If the `default_preprocessing_hooks` was used, the copyright comment could be at the beginning and still show the issue.

From my debugging, it seems like the line directive was failing to be emitted because it would errantly fall into this portion of the if-block in cpp_iterator: https://github.com/boostorg/wave/blob/develop/include/boost/wave/util/cpp_iterator.hpp#L786

Here's the output for the `t_2_032.cpp` test that I added prior to the change:
```

t_2_032_001a
#line 6 "t_2_032_001.hpp"
t_2_032_001b

t_2_032_001c

t_2_032_002

t_2_032_004
#line 2 "t_2_032_003.hpp"
t_2_032_003
#line 16 "t_2_032.cpp"
t_2_032_a

t_2_032_b
```
And here's the output after this change:
```
#line 2 "t_2_032_001.hpp"
t_2_032_001a
#line 6 "t_2_032_001.hpp"
t_2_032_001b

t_2_032_001c
#line 2 "t_2_032_002.hpp"
t_2_032_002
#line 2 "t_2_032_004.hpp"
t_2_032_004
#line 2 "t_2_032_003.hpp"
t_2_032_003
#line 16 "t_2_032.cpp"
t_2_032_a

t_2_032_b
```

PR #226, opened 2024-09-25
Line directive emitted in middle of multiline string

njnobles:
Attempts to address https://github.com/boostorg/wave/issues/225

PR reverts the changes to cpp_iterator.hpp that were introduced in https://github.com/boostorg/wave/pull/140

### DIFF
--- a/include/boost/wave/util/cpp_iterator.hpp
+++ b/include/boost/wave/util/cpp_iterator.hpp
@@ -327,12 +327,10 @@ protected:
     void on_define(parse_node_type const &node);
     void on_undefine(lexer_type const &it);
 
-    void on_ifdef(result_type const& found_directive,
-        typename parse_tree_type::const_iterator const &begin,
-        typename parse_tree_type::const_iterator const &end);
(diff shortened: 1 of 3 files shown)


jefftrull:
Unfortunately this also reverts the test :) Regardless, there's a part of #140 that I like and would want to preserve anyway:  the consolidated handling of conditional directives. Here's what I'd like to see in a PR:

1. Preservation of the existing test
2. Additional test coverage for bug #225
3. A targeted fix driven by understanding of what #140 broke and why

Any interest in working on item 2?

jefftrull:
Sorry I misunderstood the test situation here, you already thoughtfully provided one as you did for #223!

I'll try to come up with my own fix that doesn't completely revert #140 since I like the refactoring part of it. #170 is the redundant commit (w.r.t this bug).

jefftrull:
Thanks for all the work you've done on this. It's far more detail than we usually get for bugs, and I appreciate it.
We're going to close this in favor of improving the lexer error checking, rejecting these kinds of constructs entirely.

PR #233, opened 2024-11-05
De-UTF8 the directory name

tjcw:
AIX does not support UTF8 characters in directory and file names. One of the files in the tests is named like this, and this causes problems unpacking BOOST on AIX.

This PR renames the directory, and adds a stage in bootstrap.sh/bootstrap.cmd to fix up the name so that non-AIX platforms can continue to run their tests.


tjcw:
https://github.com/boostorg/boost/pull/978 is the PR that changes bootstrap.sh and bootstrap.cmd

tjcw:
https://github.com/bfgroup/b2/pull/421 is the PR in the reposistory for `b2` changes. Please refer to this rather than to boostorg/boost#978 .

tjcw:
After rechecking, it looks like AIX does not need this patch. Closing.

PR #235, opened 2024-12-06
Fix Github Actions

Flamefire:


jefftrull:
Can you briefly explain these changes? I know we were wrestling with node versions and support for old compilers last year. This PR reverts one of my workarounds/hacks for that issue. How does this one deal with the problem? Thanks.

Flamefire:
Sure. This was part of a bulk change over almost 40 repos, hence I've been very brief on description. In short:

macos-12 runners are dropped by github and jobs won't start anymore. at least macos-13 needs to be used

Containers need to be run on the latest ubuntu runner to avoid trouble, e.g. github removing the host runner os or IIRC some stuff wasn't supported on older OS.

Support for node 16 was fully dropped by GitHub --> @v3 actions don't work anymore and `$ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION` has no effect. So I removed the variable and updated to @v4 actions which require node 20. Github bind-mounts a node20 installation into the container which is incompatible with the glibc of ubuntu 16/18 and due to that mount an installation in the docker image won't work.

This solution downloads an unofficial node 20 built against the correct/older glibc and creates 2 mount points such that we can put that node onto the host runner and mount that instead of what github gives us using the idea from https://github.com/dixyes/ghactionsplay/blob/bf889ee2fed8dcd72cc8e32bfd4789be177c958f/.github/workflows/glibc217node20.yml
To avoid overloading the unofficial-builds.nodejs.org server we use a copy hosted at boost.io

Does that answer your questions or do you need more details?

jefftrull:
This fix supersedes the previous hack in #217 thanks @Flamefire

jefftrull:
Finally getting around to merging into master and I wondered if you had any insights on the CI failure. Clang, so doesn't seem related to this PR but thought I would ask https://github.com/boostorg/wave/actions/runs/14656737135

Flamefire:
See the message at the bottom of the run:
> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. 

So the failure is "intended".

I opened https://github.com/boostorg/wave/pull/239 for an update.

jefftrull:
Thanks again for your many contributions!

PR #236, opened 2025-02-28
Avoid including deprecated atomic_count.hpp header

Lastique:
`boost/detail/atomic_count.hpp` was moved to `boost/smart_ptr/detail` and the old header is deprecated and emits warnings.


jefftrull:
Thanks!

Lastique:
Looks like this didn't make it into the 1.88.0 release. Please, merge to master.

PR #239, opened 2025-04-25
GHA: Use Ubuntu 24.04 for clang test

Flamefire:
GitHub removes the Ubuntu 20 runners

Currently there are random intended failures (brownout phase):
> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. 

As the Clang version doesn't seem to matter just update to the latest OS


jefftrull:
Thanks so much for your continued help

PR #244, opened 2025-11-13
Implement support for digit separators in preprocessor expressions

jefftrull:
Integers with digit separators are already recognized as literal tokens. This commit adds proper interpretation of them as integers for use in expressions, e.g. with `#if`, by:

1. Adding digit separators to the RE2C code that handles pp-numbers,
2. Filtering out digit separators before extracting numeric values, and
3. Adding appropriate unit tests


jefftrull:
FYI @APokorny

PR #247, opened 2026-02-09
fix(macromap): correct return value in expand_predefined_macro

### LINKED ISSUE #246: Predefined macros (`__FILE__`, `__LINE__`, `__INCLUDE_LEVEL__`) are expanded twice when `expanding_object_like_macro` returns true
When a custom `expanding_object_like_macro` hook is defined and returns true, special predefined macros such as `__FILE__`, `__LINE__`, and `__INCLUDE_LEVEL__` are emitted twice in the output.

## Minimum reproduction
```cpp
#include <boost/wave.hpp>
#include <boost/wave/cpp_context.hpp>
#include <iostream>
#include <string>

struct custom_hooks : boost::wave::context_policies::default_preprocessing_hooks {
    template <typename ContextT, typename TokenT, typename ContainerT>
    bool expanding_object_like_macro(ContextT const&, TokenT const&, ContainerT const&,
                                     TokenT const&) {
        return true;
    }
};

int main() {
    std::string input =
        R"(
int main() {
     __LINE__;
}
)";

    using namespace boost::wave;

    using token_type = cpplexer::lex_token<>;
    using lexer_type = cpplexer::lex_iterator<token_type>;

    using context_type = context<std::string::const_iterator, lexer_type,
                                 iteration_context_policies::load_file_to_string, custom_hooks>;

    context_type ctx(input.begin(), input.end(), "test.cpp");

    for (auto it = ctx.begin(); it != ctx.end(); ++it) {
        if (!it->is_valid()) continue;
        std::cout << it->get_value();
    }
    std::cout << std::endl;
}
```
## Expected output
```cpp
int main() {
    __LINE__;
}
```
## Actual output
```cpp
int main() {
    __LINE__ __LINE__;
}
```

## Analysis

The issue appears to be caused by an inconsistency between
[`expand_predefined_macro`](https://github.com/boostorg/wave/blob/boost-1.90.0/include/boost/wave/util/cpp_macromap.hpp#L1671) and its caller [`expand_macro`](https://github.com/boostorg/wave/blob/boost-1.90.0/include/boost/wave/util/cpp_macromap.hpp#L1445).

In `expand_predefined_macro`:
```cpp
if (ctx.get_hooks().expanding_object_like_macro(
        ctx.derived(), deftoken, ContainerT(), curr_token))
{
    // do not expand this macro, just copy the whole sequence
    expanded.push_back(curr_token);
    return false;   // no further preprocessing required
}
```
Here, when `expanding_object_like_macro` returns true, the function pushes the predefined macro
token into `expanded` and returns **false**.

However, in `expand_macro`, macro processing is only stopped when
`expand_predefined_macro` returns **true**:
```cpp
// try to expand a predefined macro (__FILE__, __LINE__ or __INCLUDE_LEVEL__)
if (expand_predefined_macro(curr_token, expanded))
    return false;
```

As a result, the predefined macro token is emitted once inside
`expand_predefined_macro`, and then emitted again by the caller,
leading to duplicated output.

sortA0329:
Fix #246

### DIFF
--- a/include/boost/wave/util/cpp_macromap.hpp
+++ b/include/boost/wave/util/cpp_macromap.hpp
@@ -1686,7 +1686,7 @@ macromap<ContextT>::expand_predefined_macro(token_type const &curr_token,
     {
         // do not expand this macro, just copy the whole sequence
         expanded.push_back(curr_token);
-        return false;           // no further preprocessing required
+        return true;           // no further preprocessing required
     }
 
     token_type replacement;
--- a/test/testwave/collect_hooks_information.hpp
+++ b/test/testwave/collect_hooks_information.hpp
@@ -106,12 +106,18 @@ class collect_hooks_information
 
 public:
     collect_hooks_information(std::string& trace)
-      : hooks_trace(trace), skipped_token_hooks(false)
+            : hooks_trace(trace), skipped_token_hooks(false),
+                object_like_macro_passthrough(false)
     {}
 
     void set_skipped_token_hooks(bool flag) 
     {
         skipped_token_hooks = flag;
+    }
+
+    void set_object_like_macro_passthrough(bool flag)
+    {
+        object_like_macro_passthrough = flag;
     }
 
     ///////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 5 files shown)


jefftrull:
LGTM and thanks for your contribution; is there a rush on this? We are currently in the merge window - it _could_ be possible to add it directly to master with permission from the release master, but I'd prefer to wait since we can't do the normal develop->master flow.

sortA0329:
There’s no rush on this. I don't need the fix immediately, so I'm happy to wait for the normal develop->master flow.

jefftrull:
Thanks for your contribution!

PR #252, opened 2026-05-01
Copy latest AppVeyor script from boost-ci and adjust to match

jefftrull:
The [Boost CI](https://github.com/boostorg/boost-ci) project has well thought out configurations for CI we should leverage. This will be the first of several PRs syncing up with that work.


jefftrull:
This PR has CI failures that appear to be due to an ongoing DDoS attack on Canonical's servers

PR #261, opened 2026-09-20
Fix aliasing _has_include with a custom macro

### LINKED ISSUE #260: Defining an alias to _has_include fails
Preprocessing code which creates an alias to `_has_include()` via `#define FOO _has_include` cannot be processed with wave.
A real world example for this is googletest, see here: https://github.com/google/googletest/blob/52eb8108c5bdec04579160ae17225d66034bd723/googletest/include/gtest/internal/gtest-port.h#L277-L281

A minimal test case for this is
```c++
#define HAS_INCLUDE __has_include
#if HAS_INCLUDE(<header>)
#endif
```

martis42:
Fixes https://github.com/boostorg/wave/issues/260

Disclaimer: This is mostly based on an analysis with an AI Agent.


jefftrull:
From my experiments (and the provided unit test) it looks like this only handles object-like macro aliases for `__has_include`. I may prefer to hold off for a unified solution.

martis42:
You are right, I missed `#if defined BUFSIZE` and misunderstood your ask for it. Will add a test case for it an look for a proper solution.

jefftrull:
I'm happy to take on a fix, also.

martis42:
You definitely know this code better. My attempt at a logic fix would either way prefer involving an AI Agent.
Did I understand you correctly in https://github.com/boostorg/wave/issues/260#issuecomment-5781390144 that no further test cases are required, since the same bug is not present for `defined`?
Then, I think you are better suited to fix this in a more generic way. Feel free to reuse this test case. If I can help you by coming up with other tests, please ping me. I am happy to help insofar I am able to 😀

jefftrull:
The additional testcase that I think is needed is for function-like macros, i.e.

```c
#define HASINCLUDE(x) __has_include(x)
#if HASINCLUDE(x)
...
#endif
```
The expansion path is likely to be different than the object-like macros `#define HASINCLUDE __has_include`.

As for `defined()` I have convinced myself that the Wave implementation is arguably conforming.

Given that I would have to carefully review any PR for correctness I am happy to write the fix. I will be glad to copy your testcase (and credit you for it). Thanks for the bug report!

